PR #2, opened 2022-09-01
Added a guide on how to emulate the Game Boy in Rust

GiuseppeDM98:


gallottino:
These changes can be approved for me, but I think that we'll be soon replace it with a more complex README talking about the project, Discord and contribute guidelines (we need it to manage this situation). About the link: you are linking to a GB guide, but we have choosen to starting from GBA. The link could still be useful too. I think that before every pull request, an issue must be opened to discuss about desidered changes. Just some thoughts about it.

PR #7, opened 2022-09-02
Add trait Cpu and minor refactor

gallottino:


.gitignore
```diff
@@ -34,3 +34,6 @@ target/
 
 # Ignore code-workspaces
 *.code-workspace
+
+# Test roms
+*.gba
```
guerinoni, line 39: No new line at the end :(

src/cpu.rs
```diff
@@ -1,45 +1,20 @@
...
-
-    fn decode(&mut self, op: Vec<u8>) -> Condition {
-        let condition = op[3] & 0x0F; // latest 4 bit (28..32)
-        println!("condition -> {:b}", condition);
-        condition.into()
-    }
-}
+    // Abstraction of what happens for every instruction in the cpu
+    fn step(&self);
+}
```
guerinoni, line 20: No new line at the end :(

src/condition.rs
```diff
@@ -21,7 +21,7 @@ pub(crate) enum Condition {
 impl From<u8> for Condition {
     fn from(item: u8) -> Self {
         match item {
-            10 => Condition::GE,
+            0xA => Condition::GE,
```
guerinoni, line 24: Good!

guerinoni (changes requested):
Only little things :) it also miss of `cargo fmt`

.gitignore
gallottino: Done

src/cpu.rs
gallottino: Added

guerinoni:
I will merge this, it should be nice for next PR rework same commit in order to avoid a commit that don't add nothing relevant, but just nitpicks :)

PR #16, opened 2022-09-04
Refactor decode and implement branch opcode

gallottino:


src/arm7tdmi.rs
```diff
@@ -29,40 +35,90 @@ impl Cpu for Arm7tdmi {
...
+        self.program_counter += 8 + offset * 4;
+        println!("PC: {:?}", self.program_counter);
+    }
+
+    fn branch_link(&mut self, op_code: u32) {
+        todo!("Branch Link")
+    }
 }
+
+#[repr(u32)]
```
guerinoni, line 98: I don't think this is needed

src/arm7tdmi.rs
```diff
@@ -29,40 +35,90 @@ impl Cpu for Arm7tdmi {
...
+    Branch = 0x0A_00_00_00,
+    BranchLink =  0x0B_00_00_00,
+}
+
+impl ArmModeInstruction {
+
+    fn get_instruction(op_code: u32) -> Result<ArmModeInstruction, Error> {
+        use ArmModeInstruction::*;
+        
+        if Self::check(Branch as u32, op_code) {
```
guerinoni, line 110: We can create a function for `u32` like `op_code.equals(Branch)` but we can improve this in separate PR

src/arm7tdmi.rs
```diff
@@ -29,40 +35,90 @@ impl Cpu for Arm7tdmi {
...
-            } else {
-                InstructionKind::Branch
-            }
-        } else {
-            todo!()
-        };
         println!("condition -> {:x}", condition);
+
+        let res_decode = ArmModeInstruction::get_instruction(op_code);
+        if res_decode.is_err() {
```
guerinoni, line 44: Probably we need to start to handle error with `Result` or directly `panic
` if we prefer

guerinoni (changes requested):
In a single commit there are lot of changes, I would love too see them split but I don't want to be pedantic.
Add other comments and please run `cargo fmt` 🙏

src/arm7tdmi.rs
gallottino: Yes, this is only a draft to manage the instructions not implemented yet.

src/arm7tdmi.rs
gallottino: Yes, we can do it.

src/arm7tdmi.rs
gallottino: Removed

PR #20, opened 2022-09-04
Improve check instruction with specific mask for every opcode

gallottino:
Before this change, code checked only the bit set to 1. It will be added a verification for every significant bit in the opcode.


gallottino:
Only a test.

Issue #9, opened 2022-09-02
Add Cartridge class with implementation of some utilities


gallottino:
In the current situation we are reading all the info about the ROM in the main. It can be possible to add a cartridge struct in order to manage all the related behavior. More info [here](https://problemkaputt.de/gbatek.htm#gbacartridges).

simo498:
I'll take this issue

PR #22, opened 2022-09-04
Refactor of the Cartridge struct

simo498:
CartridgeHeader is now in the "Cartridge" file along with the "Cartridge" struct, which is respondible to take raw data and turn it in header fields and ROM (most of the fields are missing at the moment)

Cartridge takes the ROM data under the "rom" field in a reference counted ptr, so the type inside the ARM7TDMI implementation of the CPU was modified


gallottino:
This pull request is referring issue #9

src/cartridge_header.rs
```diff
@@ -1,62 +0,0 @@
-pub(crate) struct CartridgeHeader {
```
gallottino, line 1: You completely deleted a initial implementation of Cartridge Header. Can you merge this file with your new implementation?

src/cartridge.rs
```diff
@@ -0,0 +1,78 @@
...
+
+impl CartridgeHeader {
+    pub fn new(data: &[u8]) -> Self {
+        Self::verify_checksum(data).expect("Invalid checksum");
+        let title = String::from_utf8(data[0x00A0..0x00AC].to_vec()).expect("Invalid UTF-8 title");
+        if !title.is_ascii() {
+            panic!("Expected an ascii sequence");
+        }
+
+        /*fetch other data here*/
```
AlessandroGrassi99, line 21: Inconsistency in comment format.

It would be better to use the format
// Comment
or for docs
/// Comment

With a space between the slashes and the first word of the comment, and the first letter capitalized.

gallottino (changes requested):
We'll provide soon a guidelines for standard commit messages. The commit message like "minor changes" must completely to be avoid. In your case you can provide a pull request with just a commit.

src/cartridge_header.rs
simo498: I just moved it in the new file

simo498: Nevermind, I realized now that the implementation was changed, maybe it was added when I rebased the branch. I'll bring changes to the new file then

src/arm7tdmi.rs
```diff
@@ -1,9 +1,9 @@
-use std::fmt::Error;
+use std::{fmt::Error, rc::Rc};
 
 use crate::{condition::Condition, cpsr::Cpsr, cpu::Cpu};
 
 pub(crate) struct Arm7tdmi {
-    rom: Vec<u8>,
+    rom: Rc<Vec<u8>>,
```
guerinoni, line 6: Not sure we need `Rc` right now, but open to proposals :)

src/cartridge.rs
```diff
@@ -0,0 +1,78 @@
+use std::{
+    fs::File,
+    io::{self, Read},
+    rc::Rc,
+};
+
+///Contains all the information about the header of the cartridge
+pub(crate) struct CartridgeHeader {
```
guerinoni, line 8: The comment here doesn't add relevant information, is better to explain what is it or the behavior or other important thing, you are saying just is a thing that contains information.

I my opinion is better to keep in separated file just because if we will add tests it will be very long to read, but let's see what other members think.

src/cartridge.rs
```diff
@@ -0,0 +1,78 @@
...
+        println!("Title = {title}");
+
+        Self { title }
+    }
+
+    pub(crate) fn title(&self) -> &str {
+        self.title.as_str()
+    }
+
+    ///Verifies the header checksum
```
guerinoni, line 32: Is this comment important? As I wrote before, explain why checksum is important things to verify or other similar but just rewrite name of the function is only adding cognitive load.

guerinoni (changes requested):

src/arm7tdmi.rs
simo498: I agree it isn't needed, it was just to separate the ROM from the CPU itself and share the variable with the cartridge, a ref is also an idea. I don't think that the CPU itself should own memory stuff directly

gallottino: Personally I like it!

guerinoni: Probably will be owned by something like `InternalMemory` or sort of, for now I think `&` is enough but let's keep this if other people like it :)

simo498:
I should have got rid of most of the problems

guerinoni:
@simo498 it looks strange because I see my changes about CI here 🤔

simo498:
> @simo498 it looks strange because I see my changes about CI here 🤔 

Yeah i know, i don't actually know why they are in my commit, maybe after pulling updates from main?

guerinoni:
> > @simo498 it looks strange because I see my changes about CI here 🤔
> 
> Yeah i know, i don't actually know why they are in my commit, maybe after pulling updates from main?

Don't know what you did, but you can re-write your single commit with only you changes with a nice commit message because "Start refactoring" doesn't means nothing to other people that read that

simo498:
> > > @simo498 it looks strange because I see my changes about CI here 🤔
> > 
> > 
> > Yeah i know, i don't actually know why they are in my commit, maybe after pulling updates from main?
> 
> Don't know what you did, but you can re-write your single commit with only you changes with a nice commit message because "Start refactoring" doesn't means nothing to other people that read that

Did it👍

PR #23, opened 2022-09-04
ARM now can execute a MOV immediate instruction with S = 0

gallottino:
MOV implementation with related test.

### DIFF
--- a/src/arm7tdmi.rs
+++ b/src/arm7tdmi.rs
@@ -1,9 +1,11 @@
-use std::fmt::Error;
+use std::{fmt::Error, vec};
 
 use crate::{condition::Condition, cpsr::Cpsr, cpu::Cpu};
 
 pub(crate) struct Arm7tdmi {
     rom: Vec<u8>,
+
+    registers: [u32; 16],
 
     program_counter: u32,
     cpsr: Cpsr,
@@ -52,6 +54,9 @@ impl Cpu for Arm7tdmi {
             BranchLink => {
                 self.branch_link(op_code);
             }
+            Mov => {
+                self.mov(op_code);
+            }
             _ => todo!("Instruction not implemented yet."),
         }
     }
@@ -71,6 +76,7 @@ impl Arm7tdmi {
         Self {
             rom,
             program_counter: 0,
+            registers: [0; 16],
             cpsr: Cpsr::default(),
         }
     }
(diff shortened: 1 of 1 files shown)


src/arm7tdmi.rs
```diff
@@ -86,12 +92,45 @@ impl Arm7tdmi {
...
 
 #[derive(Debug, PartialEq, Eq)]
 pub(crate) enum ArmModeInstruction {
     Branch = 0x0A_00_00_00,
     BranchLink = 0x0B_00_00_00,
+
+    /// 27-26 must be 0b00
+    /// 24-21 must be 0b1101
+    /// 19-16 must be 0b0000
+    Mov = 0x01_A0_00_00,
```
gallottino, line 133: As the Data Processing instructions are very similar it can be refactor in a future

src/arm7tdmi.rs
```diff
@@ -86,12 +92,45 @@ impl Arm7tdmi {
...
+            let immediate_value = op_code & 0x00_00_00_FF;
+            println!("value: {:?}", immediate_value);
+
+            // the instruction is MOV RD, immediate_value
+            self.registers[rd as usize] = immediate_value;
+        } else {
+            todo!("Not implemented yet.");
+        }
+
+        self.program_counter += 4;
```
gallottino, line 121: I think that this is bad. If an instruction is `MOV PC, offset` it will be overwritten. I'll add a test to check also this behavior.

guerinoni, line 121: Instead of 4 let's use `OPCODE_ARM_SIZE`

src/arm7tdmi.rs
```diff
@@ -103,6 +142,9 @@ impl ArmModeInstruction {
         }
         if Self::check(BranchLink, op_code) {
             return Ok(BranchLink);
+        }
+        if Self::check(Mov, op_code) {
+            return Ok(Mov);
```
guerinoni, line 147: We can remove `return`

src/arm7tdmi.rs
```diff
@@ -132,4 +175,25 @@ mod tests {
...
+    fn check_mov_rx_immediate() {
+        // MOV R0, 0
+        let mut opcode: u32 = 0b11100011101000000000000000000000;
+
+        // MOV Rx,x
+        let mut cpu = Arm7tdmi::new(vec![]);
+        for rx in 0..16u32 {
+            let change_op = rx << 12;
+            opcode = (opcode & 0xFF_FF_0F_FF) + change_op;
+            opcode = (opcode & 0xFF_FF_FF_00) + rx;
```
guerinoni, line 189: I understand this test more or less but is not super clear the intention, there magic number like `0xFF_FF_0F_FF` and `0xFF_FF_FF_00` and operation with `rx` that goes `0..15` but the test is solid

gallottino: The intention is to modify the opcode resulting a cycle like this:
```
0 -> MOV R0, 0,
1 -> MOV R1, 1
...
15 -> MOV R15, 15
```
I'll add some comments above the magic numbers

src/arm7tdmi.rs
gallottino: Not sure about it. I left a comment in the code.

src/arm7tdmi.rs
guerinoni: Yes I got it, but that is the internal opcode, not "the main", is the opcode that regards only `MOV`, sure is right but coming from other context more clear code avoid cognitive load for understand this every time :D

guerinoni (approved):
LGTM

PR #29, opened 2022-09-06
Add Data Processing (ALU) Core

AlessandroGrassi99:
I implemented [this](https://rust-console.github.io/gbatek-gbaonly/#--arm-opcodes-data-processing-alu) (the chunk with ">" on the left)
```
  Bit    Expl.
  31-28  Condition
  27-26  Must be 00b for this instruction
  25     I - Immediate 2nd Operand Flag (0=Register, 1=Immediate)
  24-21  Opcode (0-Fh)               ;*=Arithmetic, otherwise Logical
           0: AND{cond}{S} Rd,Rn,Op2    ;AND logical       Rd = Rn AND Op2
           1: EOR{cond}{S} Rd,Rn,Op2    ;XOR logical       Rd = Rn XOR Op2
           2: SUB{cond}{S} Rd,Rn,Op2 ;* ;subtract          Rd = Rn-Op2
           3: RSB{cond}{S} Rd,Rn,Op2 ;* ;subtract reversed Rd = Op2-Rn
           4: ADD{cond}{S} Rd,Rn,Op2 ;* ;add               Rd = Rn+Op2
           5: ADC{cond}{S} Rd,Rn,Op2 ;* ;add with carry    Rd = Rn+Op2+Cy
           6: SBC{cond}{S} Rd,Rn,Op2 ;* ;sub with carry    Rd = Rn-Op2+Cy-1
           7: RSC{cond}{S} Rd,Rn,Op2 ;* ;sub cy. reversed  Rd = Op2-Rn+Cy-1
           8: TST{cond}{P}    Rn,Op2    ;test            Void = Rn AND Op2
           9: TEQ{cond}{P}    Rn,Op2    ;test exclusive  Void = Rn XOR Op2
           A: CMP{cond}{P}    Rn,Op2 ;* ;compare         Void = Rn-Op2
           B: CMN{cond}{P}    Rn,Op2 ;* ;compare neg.    Void = Rn+Op2
           C: ORR{cond}{S} Rd,Rn,Op2    ;OR logical        Rd = Rn OR Op2
           D: MOV{cond}{S} Rd,Op2       ;move              Rd = Op2
           E: BIC{cond}{S} Rd,Rn,Op2    ;bit clear         Rd = Rn AND NOT Op2
           F: MVN{cond}{S} Rd,Op2       ;not               Rd = NOT Op2

>   20     S - Set Condition Codes (0=No, 1=Yes) (Must be 1 for opcode 8-B)
>   19-16  Rn - 1st Operand Register (R0..R15) (including PC=R15)
>               Must be 0000b for MOV/MVN.
>   15-12  Rd - Destination Register (R0..R15) (including PC=R15)
>               Must be 0000b (or 1111b) for CMP/CMN/TST/TEQ{P}.
>   When above Bit 25 I=0 (Register as 2nd Operand)
>     When below Bit 4 R=0 - Shift by Immediate
>       11-7   Is - Shift amount   (1-31, 0=Special/See below)
>     When below Bit 4 R=1 - Shift by Register
>       11-8   Rs - Shift register (R0-R14) - only lower 8bit 0-255 used
>       7      Reserved, must be zero  (otherwise multiply or undefined opcode)
>     6-5    Shift Type (0=LSL, 1=LSR, 2=ASR, 3=ROR)
>     4      R - Shift by Register Flag (0=Immediate, 1=Register)
>     3-0    Rm - 2nd Operand Register (R0..R15) (including PC=R15)
>   When above Bit 25 I=1 (Immediate as 2nd Operand)
>     11-8   Is - ROR-Shift applied to nn (0-30, in steps of 2)
>     7-0    nn - 2nd Operand Unsigned 8bit Immediate

```

### DIFF
--- a/src/alu_instruction.rs
+++ b/src/alu_instruction.rs
@@ -1,0 +1,43 @@
+pub(crate) enum ArmModeAluInstruction {
+    AND = 0x0,
+    EOR = 0x1,
+    SUB = 0x2,
+    RSB = 0x3,
+    ADD = 0x4,
+    ADC = 0x5,
+    SBC = 0x6,
+    RSC = 0x7,
+    TST = 0x8,
+    TEQ = 0x9,
+    CMP = 0xA,
+    CMN = 0xB,
+    ORR = 0xC,
+    MOV = 0xD,
+    BIC = 0xE,
+    MVN = 0xF,
+}
+
+impl From<u8> for ArmModeAluInstruction {
+    fn from(alu_opcode: u8) -> Self {
+        use ArmModeAluInstruction::*;
+        match alu_opcode {
+            0x0 => AND,
+            0x1 => EOR,
+            0x2 => SUB,
+            0x3 => RSB,
+            0x4 => ADD,
+            0x5 => ADC,
+            0x6 => SBC,
+            0x7 => RSC,
+            0x8 => TST,
+            0x9 => TEQ,
+            0xA => CMP,
+            0xB => CMN,
+            0xC => ORR,
+            0xD => MOV,
+            0xE => BIC,
+            0xF => MVN,
+            _ => unreachable!(),
+        }
+    }
+}
--- a/src/arm7tdmi.rs
+++ b/src/arm7tdmi.rs
@@ -159,13 +159,37 @@ impl Arm7tdmi {
                                 }
                                 _ => unreachable!(),
                             },
-                            is => op2 <<= is,
+
+                            is => {
+                                match shift_type {
+                                    // Logical Shift Left
+                                    0 => op2 <<= is,
+                                    // Logical Shift Right
+                                    1 => op2 >>= is,
+                                    // Arithmetic Shift Right
+                                    2 => op2 = ((op2 as i32) >> is) as u32, // TODO: Review rust arithmetic shift right
+                                    // Rotate Right
+                                    3 => op2 = op2.rotate_right(is as u32),
+                                    _ => unreachable!(),
+                                }
+                            }
                         };
                     }
                     /// Shift by register
                     1 => {
                         let rs = ((op_code & 0x00_00_0F_00) >> 8) as u8;
-                        op2 <<= self.registers[rs as usize] & 0x00_00_00_FF;
+                        let shift_value = self.registers[rs as usize] & 0x00_00_00_FF;
+                        match shift_type {
+                            // Logical Shift Left
+                            0 => op2 <<= shift_value,
+                            // Logical Shift Right
+                            1 => op2 >>= shift_value,
+                            // Arithmetic Shift Right
+                            2 => op2 = ((op2 as i32) >> shift_value) as u32, // TODO: Review rust arithmetic shift right
+                            // Rotate Right
+                            3 => op2 = op2.rotate_right(shift_value as u32),
+                            _ => unreachable!(),
+                        };
                     }
                     _ => unreachable!(),
                 };
--- a/src/arm7tdmi.rs
+++ b/src/arm7tdmi.rs
@@ -96,13 +96,13 @@ impl Arm7tdmi {
     }
 
     fn data_processing(&mut self, op_code: u32) {
-        /// bit [25] is I = Immediate Flag
+        // bit [25] is I = Immediate Flag
         let i = ((op_code & 0x02_00_00_00) >> 25) as u8;
-        /// bit [24-21]
+        // bits [24-21]
         let alu_opcode = ((op_code & 0x01_E0_00_00) >> 25) as u8;
-        /// bit [20] is sets condition codes
+        // bit [20] is sets condition codes
         let s = ((op_code & 0x00_10_00_00) >> 20) as u8;
-        /// bits [15-12] are the Rd
+        // bits [15-12] are the Rd
         let rd = ((op_code & 0x00_00_F0_00) >> 12) as u8;
 
         let op2 = match i {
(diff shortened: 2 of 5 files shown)


AlessandroGrassi99:
I forgot to run the tests (and they failed :'( ), what should I do?

gallottino:
Can you fix it?

src/cpsr.rs
```diff
@@ -18,6 +18,14 @@ impl Cpsr {
         self.0 & 0x8000 != 0
     }
 
+    pub(crate) fn set_signed(&mut self) {
+        self.0 |= 0b1000_0000_0000_0000_0000_0000_0000_0000;
+    }
+
+    pub(crate) fn set_not_signed(&mut self) {
+        self.0 &= 0b0111_1111_1111_1111_1111_1111_1111_1111;
+    }
```
guerinoni, line 27: I don't know, probably is more clear `set_sign_flag(true|false)` but I don't have strong opinion now

src/arm7tdmi.rs
```diff
@@ -93,82 +95,113 @@ impl Arm7tdmi {
...
-        if op_code & 0x00_08_00_00 > 0 {
-            todo!("Condition set")
-        }
-
-        if immediate {
-            // bits [7-0] are the immediate value
-            let immediate_value = op_code & 0x00_00_00_FF;
-            println!("value: {:?}", immediate_value);
+    fn data_processing(&mut self, op_code: u32) {
+        /// bit [25] is I = Immediate Flag
```
guerinoni, line 99: All these commented lines should be `//`

src/arm7tdmi.rs
```diff
@@ -93,82 +95,113 @@ impl Arm7tdmi {
...
+                // 2nd Operand Register (R0..R15) (including PC=R15)
+                let mut op2 = ((op_code & 0x00_00_00_0F) >> 8);
+
+                match r {
+                    /// Shift by amount
+                    0 => {
+                        /// Shift amount
+                        let is = ((op_code & 0x00_00_07_80) >> 7) as u8;
+                        match is {
+                            0 => match shift_type {
```
guerinoni, line 124: I'm asking if is better to split the nested match for readability at least but also for testing

AlessandroGrassi99:
Fixed! :)

src/alu_instruction.rs
```diff
@@ -0,0 +1,43 @@
...
+            0x7 => RSC,
+            0x8 => TST,
+            0x9 => TEQ,
+            0xA => CMP,
+            0xB => CMN,
+            0xC => ORR,
+            0xD => MOV,
+            0xE => BIC,
+            0xF => MVN,
+            _ => unreachable!(),
```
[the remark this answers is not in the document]
AlessandroGrassi99: Unfortunately, u8 goes up to `0xFF`, so it is not possible for me to cover all cases. 

Btw I wanted to discuss that from my point of view it is better to implement `unreachable!()` than a `try_into().expect()` where possible

src/instruction.rs
```diff
@@ -0,0 +1,47 @@
...
+        (Self::get_mask(&instruction_type) & op_code) == instruction_type as u32
+    }
+
+    fn get_mask(instruction_type: &ArmModeInstruction) -> u32 {
+        use ArmModeInstruction::*;
+
+        match instruction_type {
+            Branch | BranchLink => 0x0F_00_00_00,
+            DataProcessing1 | DataProcessing2 => 0x0E_00_00_10,
+            DataProcessing3 => 0x0E_00_00_00,
```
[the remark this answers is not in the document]
AlessandroGrassi99: Yes, this is the mask, I need to check bits 27 through 25. 

From 27 to 26 I have to check that they are always 0, for the 25th bit it depends, for DataProcessing1 and DataProcessing2 it is zero, for DataProcessing3 it is 1 

```
  |..3 ..................2 ..................1 ..................0|
  |1_0_9_8_7_6_5_4_3_2_1_0_9_8_7_6_5_4_3_2_1_0_9_8_7_6_5_4_3_2_1_0|
  |_Cond__|0_0_0|___Op__|S|__Rn___|__Rd___|__Shift__|Typ|0|__Rm___| DataProc
  |_Cond__|0_0_0|___Op__|S|__Rn___|__Rd___|__Rs___|0|Typ|1|__Rm___| DataProc
  |_Cond__|0_0_1|___Op__|S|__Rn___|__Rd___|_Shift_|___Immediate___| DataProc
```

src/cpsr.rs
AlessandroGrassi99: I admit that this choice of mine is based on what I studied on Clean Code ahah

image

src/arm7tdmi.rs
AlessandroGrassi99: Fixed

src/alu_instruction.rs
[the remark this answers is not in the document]
guerinoni: For now we can keep as is, there are 2 different kind of thoughts, handling errors or fail!
I'm for handling errors in nice way but in this case we are in such situation that will never can fail, I hope, because if something panic probably we are missing some implementation...
If some instruction or conversion can fail we should handle in clean way, but now I don't have a strong answer

src/instruction.rs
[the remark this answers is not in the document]
guerinoni: Ok then how `0x0E_XX_XX_XX` can have 27-26 at 0?

src/cpsr.rs
guerinoni: I'm aware of this but in this case we are treating single bit, we don't handle the boolean in the function for doing different things... But as I wrote, I'm not sure what is the most clean and readable API here

src/instruction.rs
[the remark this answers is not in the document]
AlessandroGrassi99: It is the mask, it does not tell us the value of the bits, but which bits we need to keep for the check.

src/arm7tdmi.rs
AlessandroGrassi99: I am aware of this, however, it is still not the full implementation, there is room for improvement

gallottino (approved):

guerinoni (approved):
LGTM

PR #34, opened 2022-09-07
Now Arm can execute BIC instruction

gallottino:
BIC instruction  can be executed.

**N.B.**
BIC test must be implemented.


gallottino:
I close it for now. I'm thinking about some changes, but for now I don't have enough time to work on it.

Issue #33, opened 2022-09-07
API to manage u32 as bits


gallottino:
Discussion [here](https://discord.com/channels/919139369774891088/1016790105417072691).

AlessandroGrassi99:
I report and translate what I wrote in the discussion:

```Rust
fn get_bit(index: u8) -> u8;
fn get_bits(positions: &[u8]) -> u32;
fn get_bits_range(range: Range<u8>) -> u32;

fn check_bit(index: u8, value: u8) -> bool;
fn check_bits(positions: &[u8], values: &[u8]) -> bool;
fn check_bits_range(range: Range<u8>, values &[u8]) -> bool;
```

Regarding the use of a `u8` type instead of a boolean on `get_bit()` (`get_bit(..) -> u8` vs `get_bit(..) -> bool`)... For me reading `if opcode.get_bit(4) { }` is much less intuitive than reading `if opcode.get_bit(4) == 1 { }`. The same issue is even more pronounced in the matches. I think it is very important to make explicit what values we are looking for.

In addition, why put `get_bits_range()` if there is `get_bits()`?

Imho it is Rust's syntax that helps in readability because if we need bits 11 to 7, the difference is `opcode.get_bits_range(11..=7)` vs `opcode.get_bits(&(11..=7).collect::<Vec<u8>>())` or `opcode.get_bits(&vec![11,10,9,8,7])`

On small ranges (`opcode.get_bits(&vec![11,10,9,8,7])`) is readable, but in instructions like Branch, where the required offset ranges from 23 to 0 is tough ahah.

FedericoBruzzone:
I think this issue could be closed :)

gallottino:
Closed with #39

PR #39, opened 2022-09-08
Add bit manipulation methods

simo498:
It adds some methods to check and set single bits. For now there aren't methods to handle multiple bits masks, and it's limited to u32 (due to the assence of a common Integer trait  in rust).

Refers to #33

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,3 +9,4 @@ glium = "0.32"
 
 [dev-dependencies]
 pretty_assertions = "1.3.0"
+rand = "0.8.5"
--- a/src/bitwise.rs
+++ b/src/bitwise.rs
@@ -1,0 +1,156 @@
+/// Contains some helper methods to manipulate bits,
+/// the index (`bit_idk`) is supposed to be from lsb to msb (right to left)
+pub(crate) trait Bits {
+    fn is_bit_on(self, bit_idx: u8) -> bool;
+    fn is_bit_off(self, bit_idx: u8) -> bool;
+    fn set_bit_on(&mut self, bit_idx: u8);
+    fn set_bit_off(&mut self, bit_idx: u8);
+    fn toggle_bit(&mut self, bit_idx: u8);
+    fn set_bit(&mut self, bit_idx: u8, value: u8);
+    fn get_bit(self, bit_idx: u8) -> u8;
+}
+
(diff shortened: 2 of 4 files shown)


simo498:
> @simo498 here I appreciate more feedback also from @AlessandroGrassi99 and @gallottino 
> But sure I prefer this kind of api not for a struct `Bitwise` but as implementation of `u32` or such... In this way you can do `my_number.toggle_bit(15)` and this will toggle the bit 15.
> I have some doubt also if is better `is_set/is_reset` but let's discuss :D

Yeah it's an idea. Another is to create a wrapper like ’Bitwise(u32)’. Anyway let's hear the opinion of other people on which implementation we should uses. I also think something more direct like u32::get_bit or whatever would be more readable

src/bitwise.rs
```diff
@@ -0,0 +1,93 @@
...
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn basic_tests() {
```
gallottino, line 65: Here I prefer to split  the tests to recognize faster a future change that broke this implementation.  You can also take this test with all implementations, but I think  also it is better to have almost one test for every functionality.

gallottino (changes requested):
I also like that you create a struct from u32 and use it to manipulate the inside values. If you can change it will be great. u32 as parameter in every function is a bit redundant and also less readable. Be careful to test every function with at least an atomic test.

simo498:
I implemented the methods to u32 directly, so they can be called by just adding `use crate::bitwise::Bits`, i also added `rand` for testing, but it can be useful in the future as well

gallottino (approved):

Cargo.toml
```diff
@@ -6,6 +6,7 @@ edition = "2021"
 [dependencies]
 egui_glium = "0.19.0"
 glium = "0.32"
+rand = "0.8.5"
```
guerinoni, line 9: Can you move this as dev-dependencies?

guerinoni (changes requested):
Looks nicer :)

Cargo.toml
simo498: Sure

guerinoni (approved):
Good :)

PR #37, opened 2022-09-08
Make conversion from `hex` to `bin`

FedericoBruzzone:
Signed-off-by: Federico Bruzzone <federico.bruzzone.i@gmail.com>

PR #40, opened 2022-09-09
CPSR: Add tests for `sign_flag` and `overflow_flag`

FedericoBruzzone:
I wrote a test for `0xAAAA` representation and another for `0xAAAAAAAA` representation.
All those made with `0xAAAA` fail and those with `0xAAAAAAAA`.

P.N.
I used the hex notation for the condition to make the error more explicit. I will convert it to binary.

### DIFF
--- a/src/cpsr.rs
+++ b/src/cpsr.rs
@@ -15,7 +15,7 @@ impl Cpsr {
     }
 
     fn signed(&self) -> bool {
-        self.0 & 0x8000 != 0
+        self.0 & 0x80000000 != 0
     }
 
     pub(crate) fn set_signed(&mut self) {
@@ -27,6 +27,25 @@ impl Cpsr {
     }
 
     fn overflow(&self) -> bool {
-        self.0 & 0x1000 != 0
+        self.0 & 0x10000000 != 0
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::cpsr::Cpsr;
+
+    #[test]
+    fn check_sign_flag() {
+        let mut cpsr: Cpsr = Cpsr(0);
+        cpsr.set_signed();
+        assert!(cpsr.signed());
+    }
+
+    #[test]
+    fn check_overflow_flag() {
+        let mut cpsr: Cpsr = Cpsr(0);
+        cpsr.0 = 0b0001_0000_0000_0000_0000_0000_0000_0000;
+        assert!(cpsr.overflow());
     }
 }


src/cpsr.rs
```diff
@@ -30,3 +30,41 @@ impl Cpsr {
...
+    use crate::cpsr::Cpsr;
+    use pretty_assertions::assert_eq;
+
+    #[test]
+    fn check_sign_flag_fail() {
+        let mut cpsr: Cpsr = Cpsr(0);
+        cpsr.set_signed();
+        let output = cpsr.signed(); // cpsr.0 & 0x8000 != 0;
+        assert_eq!(output, false); // FAIL!
+    }
```
guerinoni, line 45:
```rust
    #[test]
    fn check_sign_flag_fail() {
        let mut cpsr: Cpsr = Cpsr(0);
        cpsr.set_signed();
        assert!(cpsr.signed());
    }
```

I expect this is failing

guerinoni (changes requested):

src/cpsr.rs
FedericoBruzzone: what you wrote is true. To pass the test I wrote 'cpsr.signed () == false' this implies that the test has failed. 

Would you like to see that the test fails?

guerinoni: You should not write something for make tests passing. The goal of testing is telling the true, then if I set signed, after that I expect to see `signed()` returning true... If the test fails, the bug is there :)

FedericoBruzzone: Perfect.
Locally I have already run the test you wrote and I can talk you that the bug is there. 

But, I fix everything and commit.

FedericoBruzzone:
As you can see the tests with `0xAAAA` representation are failed (`overflow_flag` ,`sign_flag`).

- Right now, would you like me to fix PR #37 with `0xAAAAAAAA` representation as it should be? 
- And about this PR do you want me to just keep the tests that pass?

src/cpsr.rs
```diff
@@ -30,3 +30,36 @@ impl Cpsr {
...
+        cpsr.set_signed();
+        assert!(cpsr.signed()); // cpsr.0 & 0x8000 != 0;
+    }
+
+    #[test]
+    fn check_sign_flag_pass() {
+        let mut cpsr: Cpsr = Cpsr(0);
+        cpsr.set_signed();
+        assert!(cpsr.0 & 0x80000000 != 0);
+    }
```
guerinoni, line 50: What is the purpose of this tests?

guerinoni:
> As you can see the tests with `0xAAAA` representation are failed (`overflow_flag` ,`sign_flag`).
> 
> * Right now, would you like me to fix PR [Make conversion from `hex` to `bin` #37](https://github.com/RIP-Comm/clementine/pull/37) with `0xAAAAAAAA` representation as it should be?
> * And about this PR do you want me to just keep the tests that pass?

Before moving on other PR I prefer to merge this fix because is important changes semantically, other refactoring of binary format are secondary in my opinion

FedericoBruzzone:
Perfect, leave the PR #37  open and I will fix it after this. 
What is missing to close this PR?

guerinoni:
> Perfect, leave the PR #37 open and I will fix it after this.
> What is missing to close this PR?

I think we can have:
- the test for checking signed (calling properly function api)
- the test for checking overflow (calling properly function api)
- the fix about signed bit

gallottino:
@guerinoni @FedericoBruzzone Also the **overflow** mask is wrong.
**_signed_** must be 0x80_00_00_00
**_overflow_** must be 0x10_00_00_00

src/cpsr.rs
gallottino: assert can be changed in this way:
`assert!(cspr.signed())`

src/cpsr.rs
```diff
@@ -30,3 +30,36 @@ impl Cpsr {
...
+        let mut cpsr: Cpsr = Cpsr(0);
+        cpsr.0 = 0b0001_0000_0000_0000_0000_0000_0000_0000;
+        assert!(cpsr.overflow()); // cpsr.0 & 0x1000 != 0;
+    }
+
+    #[test]
+    fn check_overflow_pass() {
+        let mut cpsr: Cpsr = Cpsr(0);
+        cpsr.0 = 0b0001_0000_0000_0000_0000_0000_0000_0000;
+        assert!(cpsr.0 & 0x10000000 != 0);
```
gallottino, line 63: Same of above:
`assert!(cspr.overflow())`

gallottino (changes requested):

FedericoBruzzone:
The purpose of test 'test_sign_flag_pass' and test 'test_overflow_flag_pass' is to show that '0xAAAAAAAA' format is correct.  

By changing to '0x80000000' and '0x10000000' I will replace with 'cpsr.signed()' and 'cpsr.overflow()'

FedericoBruzzone:
I have already done the test to verify that overflow is wrong (0x1000) @gallottino

gallottino:
I'm talking about the bug fixs. I think It is better to fix them in this PR before merge these change. @FedericoBruzzone

FedericoBruzzone:
Of course, what you say I will definitely do tomorrow!

And I will convert 0x80000000 and 0x10000000 in binary after the merge in #37.

guerinoni (approved):
LGTM 🏆

gallottino (approved):

PR #42, opened 2022-09-11
ARM: Implement `teq` `ALU` instruction

guerinoni:
This don't cover all case because we are not taking care of changing all CPU register and other marginal stuff,
but it's a start :).

### DIFF
--- a/src/arm7tdmi.rs
+++ b/src/arm7tdmi.rs
@@ -1,6 +1,7 @@
 use std::convert::TryInto;
 
 use crate::alu_instruction::ArmModeAluInstruction;
+use crate::bitwise::Bits;
 use crate::instruction::ArmModeInstruction;
 use crate::{condition::Condition, cpsr::Cpsr, cpu::Cpu};
 
@@ -141,12 +142,7 @@ impl Arm7tdmi {
                                 1 => {
                                     // TODO: It's better to implement the logical instruction in order to execute directly LSR#0?
                                     let rm = self.registers[op2 as usize];
-                                    match (rm & 0b1000_0000_0000_0000_0000_0000_0000_0000) >> 31 {
-                                        1 => self.cpsr.set_signed(),
-                                        0 => self.cpsr.set_not_signed(),
-                                        _ => unreachable!(),
-                                    }
-
+                                    self.cpsr.set_signed(rm.get_bit(31));
                                     op2 = 0;
                                 }
                                 // ASR#0: Interpreted as ASR#32, ie. Op2 and C are filled by Bit 31 of Rm.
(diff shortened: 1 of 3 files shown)


src/arm7tdmi.rs
```diff
@@ -227,13 +234,29 @@ impl Arm7tdmi {
 
         match ArmModeAluInstruction::from(alu_opcode) {
             ArmModeAluInstruction::Mov => self.mov(rd as usize, op2),
-            _ => todo!(),
+            ArmModeAluInstruction::Teq => {
+                let val = self.teq(rn, op2);
+                self.set_registers(rd, val);
```
gallottino, line 239: Rn shouldn't be changed. I think this opcode is only about flags.
[here](https://www.problemkaputt.de/gbatek.htm#armopcodesdataprocessingalu) is:
`Void = Rn XOR Op2`

In other instructions there is something like this:
`Rd = Rn OR Op2`

src/arm7tdmi.rs
```diff
@@ -277,4 +300,18 @@ mod tests {
...
+        let op_code: u32 = 0b1110_0001_0010_1001_0011_0000_0000_0000;
+        let mut cpu = Arm7tdmi::new(vec![]);
+
+        let (_, instruction) = cpu.decode(op_code);
+        assert_eq!(instruction, ArmModeInstruction::DataProcessing1);
+
+        cpu.execute(op_code, instruction);
+        let rn = 0b0011;
+        let rd_xor_op2 = 0b1001;
+        assert_eq!(cpu.registers[rn], rd_xor_op2);
```
gallottino, line 315: For the above reason this test should be different.

gallottino (changes requested):

src/arm7tdmi.rs
guerinoni: Yes, indeed is not modified

src/arm7tdmi.rs
guerinoni: The right name is `rd` I'm updating it

src/arm7tdmi.rs
gallottino: Sorry, I would to say Rd.

guerinoni: What this operator should do with the result of that XOR?

gallottino: I think  this opcode changes only the flags in CPSR

guerinoni: In which way? You mean something like `cpsr = (result of XOR)` (since cpsr is a u32...)

guerinoni: I figured out the right way :)

gallottino (approved):

PR #43, opened 2022-09-12
Move checksum validation into its function

### LINKED ISSUE #28: Check if cartridge checksum is correct
It should be in `cartridge_header.rs`. 

Specific in the function:
``` rust
fn extract_complement_check(data: &[u8]) -> [u8; 1] {
        // TODO: this issue
        data[0x0BD..=0x0BD]
            .try_into()
            .expect("extracting complement check")
    }
```

Reference [here](https://problemkaputt.de/gbatek.htm#gbacartridgeheader) at `Complement check` header

simo498:
Should close #28

### DIFF
--- a/src/cartridge_header.rs
+++ b/src/cartridge_header.rs
@@ -10,7 +10,7 @@ pub(crate) struct CartridgeHeader {
     pub(crate) device_type: [u8; 1],
     pub(crate) reserved_area_1: [u8; 7],
     pub(crate) software_version: [u8; 1],
-    pub(crate) complement_check: [u8; 1],
+    pub(crate) complement_check: u8,
     pub(crate) reserved_area_2: [u8; 2],
     pub(crate) ram_entry_point: [u8; 4],
     pub(crate) boot_mode: [u8; 1],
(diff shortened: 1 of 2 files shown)


FedericoBruzzone:
Can you write a test to see if everything is ok?

simo498:
> Can you write a test to see if everything is ok?

Let's open an issue for that!

guerinoni:
The behavior is different, because before of this changes the function it returns a `Result` that enable future better error handling, right now we have a panic instead... 
I don't know if this is testable, sure we can add a unit test in this PR if we want but we can't add to the repo a real ROM then...

simo498:
I guess we can rethink of the whole function to give back a result if it's needed, i removed it because the initializing function of the cartridge was meant to be infallible (since it's useless to catch an error if the cartridge is invalid)

guerinoni:
> I guess we can rethink of the whole function to give back a result if it's needed, i removed it because the initializing function of the cartridge was meant to be infallible (since it's useless to catch an error if the cartridge is invalid)

I know, it was there for future improvement, keep it and add a Result on `new` function, in this way we can handle correctly from the caller, one day we will add better error handling.

I guess some kind of error are not-acceptable error, like reading in wrong area of memory, and there is a bug from our implementation, other things like checksum or nintendo logo are only nitpicks and we can handle in other way for add a workaround in future, like "checksum is wrong, start anyway?" Because these kind of check were studied for avoid hacking cartridge or fake clone... We don't really care of this :D

simo498:
Ok, that's fine 👍

guerinoni (approved):

PR #44, opened 2022-09-12
CPSR: Add all `ARM condition filed` and finished `CPSR`

### LINKED ISSUE #8: Add condition behavior in CPSR register
In cpsr.rs file must be added all the behavior based on the condition parameter of the OPCODE.
```
pub(crate) fn can_execute(&self, cond: Condition) -> bool {
      match cond {
          Condition::GE => self.signed() == self.overflow(),
          // TODO: complete pattern matching for every condition
          _ => todo!(),
      }
}
```
When an instruction has been decoded the ARM cpu executes it only if the flags in the CSPR match the condition. More information [here](https://problemkaputt.de/gbatek.htm#armcpuflagsconditionfieldcond)

FedericoBruzzone:
Closes #8 

P.N.
The only part missing is `saved_program_status_register (SPSR_<mode>)` of the CPSR

Sign-off-by: Federico Bruzzone <federico.bruzzone.i@gmail.com>

### DIFF
--- a/src/condition.rs
+++ b/src/condition.rs
@@ -41,3 +41,37 @@ impl From<u8> for Condition {
         }
     }
 }
+
+/// Control bits condition.
+pub enum ModeBits {
+    OldUser = 0x00,
+    OldFiq = 0x01,
+    OldIrq = 0x02,
+    OldSupervisor = 0x03,
+    User = 0x10,
+    Fiq = 0x11,
+    Irq = 0x12,
+    Supervisor = 0x13,
+    Abort = 0x17,
+    Undefined = 0x1B,
+    System = 0x1F,
+}
+
+impl From<u8> for ModeBits {
+    fn from(item: u8) -> Self {
+        match item {
+            0x00 => Self::OldUser,
+            0x01 => Self::OldFiq,
+            0x02 => Self::OldIrq,
+            0x03 => Self::OldSupervisor,
+            0x10 => Self::User,
+            0x11 => Self::Fiq,
+            0x12 => Self::Irq,
+            0x13 => Self::Supervisor,
+            0x17 => Self::Abort,
+            0x1B => Self::Undefined,
+            0x1F => Self::System,
+            _ => unreachable!(),
+        }
+    }
+}
--- a/src/cpsr.rs
+++ b/src/cpsr.rs
@@ -1,3 +1,5 @@
+#[cfg(test)] // TODO: remove cfg when this API will be used at least one in prod code.
+use crate::condition::ModeBits;
 use crate::{bitwise::Bits, condition::Condition};
 
 /// Current Program Status Register.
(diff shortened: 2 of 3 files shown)


src/condition.rs
```diff
@@ -41,3 +41,37 @@ impl From<u8> for Condition {
         }
     }
 }
+
+// Control bits condition
```
guerinoni, line 45: Is this documentation? Because instead just remove it, doesn't add any relevant information

src/condition.rs
```diff
@@ -41,3 +41,37 @@ impl From<u8> for Condition {
...
+pub(crate) enum ModeBits {
+    OldUser = 0x00,
```
guerinoni, line 47: Binary or Hex representation?

src/cpsr.rs
```diff
@@ -1,51 +1,293 @@
 use crate::condition::Condition;
+use crate::condition::ModeBits;
 
-/// Current Program Status Register.
+/// Current Program Status Register
```
guerinoni, line 4: Keep the `.`
In the documentation is important to respect the grammar.

src/cpsr.rs
```diff
@@ -1,51 +1,293 @@
...
+            PL => !self.sign_flag(),                        // Positive or zero (N=0)
+            VS => self.overflow_flag(),                     // Overflow (V=1)
+            VC => !self.overflow_flag(),                    // No overflow (V=0)
+            HI => self.carry_flag() && !self.zero_flag(),   // Unsigned higher (C=1 and Z=0)
+            LS => !self.carry_flag() || self.zero_flag(),   // Unsigned lower or same (C=0 and Z=1)
+            GE => self.sign_flag() == self.overflow_flag(), // Greater or equal (N=V)
+            LT => self.sign_flag() != self.overflow_flag(), // Less than (N<>V)
+            GT => !self.zero_flag() && (self.sign_flag() == self.overflow_flag()), // Greater than (Z=0 and N=V)
+            LE => self.zero_flag() && (self.sign_flag() != self.overflow_flag()), // Less or equal (Z=1 or N<>V)
+            AL => true,  // always (the "AL" suffix can be omitted)
```
guerinoni, line 26: Missing capital `A`

src/cpsr.rs
```diff
@@ -1,51 +1,293 @@
...
+            NV => false, //never (ARMv1, v2 only) (Reserved ARMv3 and up)
```
guerinoni, line 27: Missing space and capital `N`

src/cpsr.rs
```diff
@@ -1,51 +1,293 @@
...
+            ModeBits::Undefined => self.0 |= 0b0000_0000_0000_0000_0000_0000_0001_1011,
+            ModeBits::System => self.0 |= 0b0000_0000_0000_0000_0000_0000_0001_1111,
+        }
     }
 }
 
 #[cfg(test)]
 mod tests {
-    use crate::cpsr::Cpsr;
+    use crate::{condition::ModeBits, cpsr::Cpsr};
```
guerinoni, line 198: Change with `super::*` is easier for unit tests

guerinoni (changes requested):
Only nitpicks... Make attention to `clippy` also :P

src/condition.rs
FedericoBruzzone: Yes, it is. I will proceed to correct it with `///`.

src/condition.rs
FedericoBruzzone: For the `condition` and `istruction` (`alu_istruction.rs` && `condition.rs` al the moment) we decided to use the `hex` notation.

But, we can still talk about it.

src/cpsr.rs
FedericoBruzzone: :)

src/cpsr.rs
FedericoBruzzone: :)

src/cpsr.rs
FedericoBruzzone: :)

src/condition.rs
guerinoni: It's ok for me!

src/cpsr.rs
```diff
@@ -1,3 +1,5 @@
+#[cfg(test)] // TODO: remove cfg when this API will be used at least one in prod code.
+use crate::condition::ModeBits;
```
guerinoni, line 2: If we need this only in test just put under `mod test`

FedericoBruzzone: I understand that maybe I've written to many lines of code.
However as you can see `ModeBits` is used at line 157, but since the function isn't used at the moment I had to mark `use crate::condition::ModeBits;` with `#[cfg(test)]`

Immagine 2022-09-13 235810

guerinoni: Ok, if it isn't used you can mark this as `#[allow(dead_code)]`... But ok I understood your goal.

FedericoBruzzone: Ok, perfect!

PR #45, opened 2022-09-13
CI: Enable more pedantic warnings on clippy

guerinoni:
In the near future we can shortcut this long command with `make` or `just`.
Another improvements is enable all warnings, but is a long run :)


FedericoBruzzone:
It's ok for me.
LGTM

PR #48, opened 2022-09-14
ARM: Add `branch_link` implementation  for `arm7tdmi`

### LINKED ISSUE #46: ARM: `branch_link` instruction must be implemented for `arm7tdmi`
Documentation [here](https://rust-console.github.io/gbatek-gbaonly/#arm-binary-opcode-format)

FedericoBruzzone:
Closes #46 

It is a first implementation.
This implementation looks like that of `branch`, but if the bit 24 is set to 1 we need save the PC in cell 14 of the `registers`.

### DIFF
--- a/src/arm7tdmi.rs
+++ b/src/arm7tdmi.rs
@@ -93,8 +93,14 @@ impl Arm7tdmi {
         println!("PC: {:?}", self.program_counter);
     }
 
-    fn branch_link(&mut self, _op_code: u32) {
-        todo!("Branch Link")
+    fn branch_link(&mut self, op_code: u32) {
+        self.registers[14] = self.program_counter.wrapping_add(4); //R14 = LR
+
+        let offset = op_code & 0b0000_0000_1111_1111_1111_1111_1111_1111;
+        println!("offset: {:?}", offset);
+
+        self.program_counter += 8 + offset * 4;
+        println!("PC: {:?}", self.program_counter);
     }
 
     fn data_processing(&mut self, op_code: u32) {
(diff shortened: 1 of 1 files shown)


FedericoBruzzone:
Closes #48

src/arm7tdmi.rs
```diff
@@ -260,6 +266,22 @@ mod tests {
...
+    #[test]
+    fn decode_branch_link() {
+        let output: Result<ArmModeInstruction, String> =
+            0b1110_1011_0000_0000_0000_0000_0111_1111.try_into();
+        assert_eq!(output, Ok(ArmModeInstruction::BranchLink));
+    }
+
+    #[test]
+    fn test_registers_14_after_branch_link() {
+        let mut output: Arm7tdmi = Arm7tdmi::new(vec![]);
```
guerinoni, line 278:
```suggestion
        let mut cpu: Arm7tdmi = Arm7tdmi::new(vec![]);
```

src/arm7tdmi.rs
```diff
@@ -93,8 +93,14 @@ impl Arm7tdmi {
...
 
-    fn branch_link(&mut self, _op_code: u32) {
-        todo!("Branch Link")
+    fn branch_link(&mut self, op_code: u32) {
+        self.registers[14] = self.program_counter;
+
+        let offset = op_code & 0b0000_0000_1111_1111_1111_1111_1111_1111;
+        println!("offset: {:?}", offset);
+
+        self.program_counter += 8 + offset * 4;
```
guerinoni, line 102: This depends on opcode at bit 24, there are 3 different implementation, please choose one to implement, this seems `B` (=0)

guerinoni (changes requested):

src/arm7tdmi.rs
FedericoBruzzone: Look at this
Immagine 2022-09-16 104912

`B` (=0) is `Branch` 
`BL` (=1) is `BranchLink`
as you can see in this implementation.

Immagine 2022-09-16 110807

The only difference is that `BL` saves `PC + 4` in `LR` (R14)

We wrote `B` and `BL` separately.
Let's talk about it if we want to write them in a function.

P.N.
Exist also `BLX`
Immagine 2022-09-16 111244

guerinoni: Then, the only missing implementation is BLX, right?

FedericoBruzzone: Yes, exactly! I have a question about this...
- What does `Opcode` = `H` mean?
I could suppose `Halfword Offset for BLX`, so we cannot implement `BLX` like `Branch` and `Branch With Link` because we cannot set 24th bit to`H` (obviously).

FedericoBruzzone: I can open an issue about this implementation.

guerinoni: Ok for now it's ok, during next interactions we will refactor this stuff in more suited way

FedericoBruzzone: I wrote in the previous comment If we want to write 'branch' and 'branch with link' in a single function. I think we should write this in a suited way, as you said.

PR #50, opened 2022-09-15
API: Add `clear` function to `bitwise`

### LINKED ISSUE #49: Add `clear` function to `bitwise.rs`
To complete API bitwise, I think we need to implement two more funcion:

- `clear_bits_msb_to_idx(&self, bit_idx: u8)` // msb: most significant bit
- `clear_bits_idx_to_0(&self, bit_idx: u8)`

FedericoBruzzone:
Closes #49 

These function complete the bitwise file.
- The first one allow us to clean (set to 0) all bits from `msg` to `idx`;
- The second one allow us to clean (set to 0) all bits from `idx` to `0`

### DIFF
--- a/src/bitwise.rs
+++ b/src/bitwise.rs
@@ -12,6 +12,7 @@ pub trait Bits {
     fn get_bit(self, bit_idx: u8) -> bool;
     fn get_bits(self, bits_range: RangeInclusive<u8>) -> u32;
     fn are_bits_on(self, bits_range: RangeInclusive<u8>) -> bool;
+    fn clear_bits(&mut self, bit_idx: u8, to: u8);
 }
 
 impl Bits for u32 {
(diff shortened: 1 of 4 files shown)


FedericoBruzzone:
Some function (in particular set functions) on `bitwise` file uses in its body for example `let mask = 0b1 << bit_idx` and then `*self &= mask`. 
Can I change it in `*self &= (0b1 << bit_idx)`?

src/bitwise.rs
```diff
@@ -57,13 +59,43 @@ impl Bits for u32 {
...
+        b.clear_bits_msb_to_idx(3);
+        println!("{}", b)
+    }
+
+    #[test]
+    fn test_clear_bits_idx_to_0() {
+        let mut b = 0b11111;
+        b.clear_bits_idx_to_0(3);
+        println!("{}", b)
+    }
```
guerinoni, line 97: Tests without assertion? 🤔

src/bitwise.rs
```diff
@@ -8,6 +8,8 @@ pub trait Bits {
     fn toggle_bit(&mut self, bit_idx: u8);
     fn set_bit(&mut self, bit_idx: u8, value: bool);
     fn get_bit(self, bit_idx: u8) -> bool;
+    fn clear_bits_msb_to_idx(&mut self, bit_idx: u8); // msb: most significant bit
+    fn clear_bits_idx_to_0(&mut self, bit_idx: u8);
```
guerinoni, line 12: Where do you we need these right now?

src/bitwise.rs
FedericoBruzzone: I'm sorry, I had a big problem afternoon and i didn't check well.

src/bitwise.rs
FedericoBruzzone: At the moment they will not be used but for the sake of completeness I have thrown them in.

gallottino (commented):
I prefer to have only one method with left_index and right_index.

FedericoBruzzone:
Perfect, I'll merge my two functions 👍

gallottino (approved):

gallottino:
> Perfect, I'll merge my two functions 👍

I'm not sure that in this way there is an improvment. I don't like that someone has forced to start from msb or lsb.
For example:


101101

I want to clear only the two centered '1'. With this method I can't. Am I missing something?

FedericoBruzzone:
I did not understand.

I fixed everything!

gallottino (approved):

gallottino (changes requested):
@FedericoBruzzone There is some problem in the history: you add a void merge from main to branch and your changes are not isolated from the rest. I think that for this pull request just a commit with "add clear_bits method in bitwise" must to appear. 

Anyway we don't need this method yet, so I don't know if it is worth working on it.

PR #53, opened 2022-09-16
Add `get_bits` method for `u32`

AlessandroGrassi99:
Implement the `get_bits` method that takes as input a RangeInclusive<u8>. 

Unfortunately, in Rust it is only possible to make ranges in ascending order, the opposite of the formatting used in the documentation. However, I think it still can greatly increase the readability of the codebase.

### DIFF
--- a/src/bitwise.rs
+++ b/src/bitwise.rs
@@ -63,8 +63,9 @@ impl Bits for u32 {
 
     fn get_bits(self, bits_range: RangeInclusive<u8>) -> u32 {
         let mut bits = 0b0;
-        for (range_idx, range_value) in bits_range.enumerate() {
-            bits |= (self.get_bit(range_value) as u32) << range_idx;
+        for (shift_value, bit_index) in bits_range.enumerate() {
+            let bit_value: Self = self.get_bit(bit_index).into();
+            bits |= bit_value << shift_value;
         }
         bits
     }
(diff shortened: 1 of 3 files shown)


src/bitwise.rs
```diff
@@ -57,6 +60,15 @@ impl Bits for u32 {
...
         self.is_bit_on(bit_idx)
     }
+
+    fn get_bits(self, bits_range: RangeInclusive<u8>) -> u32 {
+        let mut bits = 0b0;
+        for (shift_value, bit_index) in bits_range.enumerate() {
+            let bit_value: Self = self.get_bit(bit_index).into();
+            bits |= bit_value << shift_value;
+        }
+        bits
```
guerinoni, line 70: We can write as, but not very important :)
```rust
        bits_range.into_iter().enumerate().fold(0, |bits, (v, i)| {
            let bit_value: Self = self.get_bit(i).into();
            bits | bit_value << v
        })
```

guerinoni (approved):
LGTM, it would be nice to see this API used in the code and not only in trait :)

src/bitwise.rs
AlessandroGrassi99: Interesting. 

I thought that maybe get_bits would also work as a default implementation of the trait.  what do you think?

guerinoni: Probably, 🤔

guerinoni (approved):
Nice

PR #57, opened 2022-09-20
Refactor shift operation

AlessandroGrassi99:
Since the shift operation is used by several instructions, I preferred to separate it from `Data Processing (ALU)`.

I made two methods, `shift` and `shift_immediate`, the first one performs special operations in case the `shift_amount` is `0`, the second one performs the shift operation directly.

### DIFF
--- a/src/arm7tdmi.rs
+++ b/src/arm7tdmi.rs
@@ -131,75 +131,19 @@ impl Arm7tdmi {
                 let mut op2 = opcode.get_bits(0..=3);
 
                 match r {
+                    // 0=Immediate, 1=Register
                     // Shift by amount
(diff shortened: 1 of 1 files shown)


FedericoBruzzone:
Looks good.

Have you thought about passing `op2` by reference?
By doing so could we avoid reassignement?

AlessandroGrassi99:
> Looks good.
> 
> Have you thought about passing `op2` by reference? By doing so could we avoid reassignement?

I thought about it, but u32 implements Copy, so it doesn't seem a noteworthy advantage to me.

guerinoni (approved):
Looks good!

PR #58, opened 2022-09-22
Improve `Single Data Transfer`

AlessandroGrassi99:


AlessandroGrassi99:
Even if we have to rethink the core operation of education, I think it doesn't affect this PR very much.

guerinoni:
Is this still valid?

PR #60, opened 2022-09-25
ARM: Implement PC within same bank of 16 register for CPU

guerinoni:
Before of this PC was a separated variable, in original design PC is only
R15 of array of registers.

Right now in `check_mov_rx_immediate` we verify immediate value from 0 to 15, is this a realistic situation?
Another question, we are advancing PC after `execution`, does this make sense?

We can also skip those question right now, just some thoughts for future work :)

### DIFF
--- a/src/arm7tdmi.rs
+++ b/src/arm7tdmi.rs
@@ -5,12 +5,39 @@ use crate::bitwise::Bits;
 use crate::instruction::ArmModeInstruction;
 use crate::{condition::Condition, cpsr::Cpsr, cpu::Cpu};
 
+/// Contains the 16 registers for the CPU, latest (R15) is special because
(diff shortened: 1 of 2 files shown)


FedericoBruzzone:
Your implementation looks good.
- Did you think about if it is better to move `Registers` implementation to a new file?
- About the second question:
  _instruction pointer_: after that the processor decodes the current instruction it finds how many bytes it occupies and 
  obviously knows how many bytes to add to the current _instruction pointer_ to reach the next instruction, and so far so good; 
  Some processor, where all instruction have the same length, can add 4 (after execution) uncoditionally, I think this is the 
  answer.

AlessandroGrassi99 (approved):

Issue #59, opened 2022-09-24
Docker support


mastrogiovanni:
Clementine could be run within a Docker container. It should be created a proper Dockerfile, documentation and some github action to build and push the container into a public registry when a PR is merged.

PR #61, opened 2022-09-26
59 docker support

mastrogiovanni:
Created a dockerfile for clementine and github action to push docker image on GHCR.
It is aimed to solve the issue #59

### DIFF
--- a/.github/workflows/publish-ghcr.yml
+++ b/.github/workflows/publish-ghcr.yml
@@ -1,4 +1,4 @@
-name: Publish Clementine on DockerHub
+name: Publish Clementine on GHCR
 
 on:
   push:
@@ -26,13 +26,13 @@ env:
 
 jobs:
   push_clementine_to_registry:
-    name: Push Clementine image to Docker Hub
+    name: Push Clementine image to GHCR
     runs-on: ubuntu-latest
     steps:
       - name: Check out the repo
         uses: actions/checkout@v3
       
-      - name: Log in to Docker Hub
+      - name: Log in to GHCR
         uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
         with:
           registry: ${{ env.REGISTRY }}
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-ghcr.yml
```diff
@@ -0,0 +1,55 @@
...
+      - "Cargo.toml"
+      - "Cargo.lock"
+  pull_request:
+    branches:
+      - 'main'
+    paths: 
+      - "src/**"
+      - "Cargo.toml"
+      - "Cargo.lock"
+  workflow_dispatch:
```
guerinoni, line 21: Probably is enough on a `tags`

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM rust:1.63 as builder
```
guerinoni, line 1: Latest release is 1.64

mastrogiovanni:
I found a criticity: Docker image is not working given insertion of UI. I guess we need:
1) found a way to allow UI to work within docker or
2) separate the UI from the rust codebase and let two modules communicate via some protocol (gRPC, HTTP..)

.github/workflows/publish-ghcr.yml
mastrogiovanni: I already tested the possibility to push into GHCR: it worked.

guerinoni:
> I found a criticity: Docker image is not working given insertion of UI. I guess we need:
> 
> 1. found a way to allow UI to work within docker or
> 2. separate the UI from the rust codebase and let two modules communicate via some protocol (gRPC, HTTP..)

Maybe the easiest is the first one, also because splitting UI it doesn't make really sense, since without UI is a useful gameboy 😓

FedericoBruzzone:
I also think we shouldn't separate the UI. 
The problem is that the container does not support it?

Dockerfile
mastrogiovanni: In general is always better to choose a specific version instead of "latest" in order to have reproducible build. However this PR is on hold.

guerinoni:
Closing for now because it's a weird way that we need to define before drafting some code

PR #62, opened 2022-09-27
BITS: Added `are_bits_on` function and implemented in ARM.

Peppe3009:
As a form of improvement of the existing functionalities I have implemented the `are_bits_on` function in order to simply check if a certein sub-sequence of bit is all set to 1.
I've made all the necessary tests and the function seems to work just fine so i decided to use it insted of the verbose code in `SingleDataTransfer`.
I also decide to not get rid of the old code because it might be useful in future debugging situations.


.gitignore
```diff
@@ -40,3 +40,6 @@ target/
 
 # Test roms
 *.gba
+
+# Ignore Visual Studio 
+.vs/
```
guerinoni, line 45: Missing new line at the end of file

src/arm7tdmi.rs
```diff
@@ -313,13 +313,14 @@ enum SingleDataTransfer {
...
-        // TODO: possible improvements
-        // - op_code.are_bits_on(31..28)
-        // - op_code.is_on(31).and(30).and(29)...
+        /*
         let must_for_pld = op_code.is_bit_on(31)
             && op_code.is_bit_on(30)
             && op_code.is_bit_on(29)
             && op_code.is_bit_on(28);
+        */
+
```
guerinoni, line 322: Remove commented code

src/bitwise.rs
```diff
@@ -69,6 +71,23 @@ impl Bits for u32 {
...
         bits
     }
+
+    /*
+        Check if the bits in a certein range are all set.
+        Return false when there is at least 1 bit which is 0, true otherwise.
+        When all bits are setted to 1, the check in the inner loop fails and true is returned.
+
+    
+    */
```
guerinoni, line 81: Change with `///` for documentation

src/bitwise.rs
```diff
@@ -69,6 +71,23 @@ impl Bits for u32 {
...
+    fn are_bits_on(self, bits_range: RangeInclusive<u8>) -> bool {
+
```
guerinoni, line 83: Remove this empty line

guerinoni (changes requested):

PR #64, opened 2022-09-28
BITS: Added `are_bits_on` function and implemented in ARM.

Peppe3009:
As a form of improvement of the existing functionalities I have implemented the `are_bits_on` function in order to simply check if a certein sub-sequence of bit is all set to 1.
I've made all the necessary tests and the function seems to work just fine so i decided to use it instead of the verbose code in `SingleDataTransfer`.
The code has also been refactored to satisfy project requirements.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,6 @@ target/
 
 # Test roms
 *.gba
+
+# Ignore Visual Studio 
+.vs/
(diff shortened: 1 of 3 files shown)


guerinoni:
Can you squash all in one commit and don't use caps lock. Something like 
`Bitwise: Introduce method `are_bits_on``

PR #67, opened 2022-09-29
Improve `CartridgeHeader`

### LINKED ISSUE #9: Add Cartridge class with implementation of some utilities

simo498:
Close #9


src/cartridge_header.rs
```diff
@@ -196,25 +157,47 @@ impl CartridgeHeader {
...
+        .wrapping_sub(0x19);
+
+    if checksum != checksum_expected {
+        return Err(format!("Expected {} but got {}", checksum_expected, checksum).into());
+    }
+
+    Ok(())
+}
+
+fn extract_str(data: &[u8]) -> Result<String, Box<dyn Error>> {
```
guerinoni, line 193: This is more like `into_ascii_str`

src/main.rs
```diff
@@ -36,16 +35,8 @@ fn main() {
...
-        Ok(d) => d,
-        Err(e) => {
-            println!("{e}");
-            std::process::exit(2);
-        }
-    };
-
-    let cartridge_header = CartridgeHeader::new(&data).unwrap();
-    println!("{}", cartridge_header.game_title);
+    let (cartridge_header, data) = CartridgeHeader::from_file(name).unwrap();
```
guerinoni, line 38: Before we were handling but right now we just unwrap...

simo498:
I fixed some stuff, like taking the array for the nintendo logo by reference, i completed removed unused fields

guerinoni (commented):
More or less looks good, I have some concerns on what improvements brings this PR... The commit is "refactor something" that doesn't really means nothing...
I'm not against to merge this, but I appreciated to have more clear what we are introducing in the main :)

simo498:
> More or less looks good, I have some concerns on what improvements brings this PR... The commit is "refactor something" that doesn't really means nothing... I'm not against to merge this, but I appreciated to have more clear what we are introducing in the main :)

It's just to make clearer some messed up code: the field's types are like [u8; 1] or [u8; 4] for opcodes, which is really unclear and bug prone when converting to u32 (like confusing little or big endian), most of methods were a bit boilerplate, and fields shouldn't be public in my opinion (they should be immutable once initialized). Anyway, I wrote some infos in the commit description

guerinoni:
Is this still valid?

PR #68, opened 2022-09-29
ARM: Implement `CMP` ALU Instruction

fedebuonco:
Straightforward implementation of the CMP instruction. Based on the previous implementation of TEQ. More test should be written.


guerinoni (commented):
Looks good, it needs a rebase

PR #69, opened 2022-09-30
Add `Opcode` struct

AlessandroGrassi99:
I think using `u32` to handle opcodes has become frustrating and confusing. Therefore, I created a structure that can handle the main functions of the opcode: 
- Condition
- Instruction
- Bits Manipulation
- Debug/Display

Unfortunately, the `Display` support for `Single Data Transfer` is missing since currently `ArmModeInstruction::DataTransfer` is used to represent multiple instructions at once. Therefore, it is necessary to divide `DataTransfer` in a manner similar 
to `DataProcessing` (1, 2, and 3). BTW, If you do not use the `Display` trait of `Opcode` in `Arm7tdmi::decode()` everything works as before.

### DIFF
--- a/src/cpsr.rs
+++ b/src/cpsr.rs
@@ -7,7 +7,7 @@ use crate::{bitwise::Bits, condition::Condition};
 pub struct Cpsr(u32);
 
 impl Cpsr {
-    pub(crate) fn can_execute(&self, cond: &Condition) -> bool {
+    pub(crate) fn can_execute(&self, cond: Condition) -> bool {
         use Condition::*;
         match cond {
             EQ => self.zero_flag(),                         // Equal (Z=1)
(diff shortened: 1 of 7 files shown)


src/cpsr.rs
```diff
@@ -7,7 +7,7 @@ use crate::{bitwise::Bits, condition::Condition};
 pub struct Cpsr(u32);
 
 impl Cpsr {
-    pub(crate) fn can_execute(&self, cond: Condition) -> bool {
+    pub(crate) fn can_execute(&self, cond: &Condition) -> bool {
```
guerinoni, line 10: Probably if we add `Copy` trait there's no need to add reference here, but no blocker from my side.

guerinoni (approved):
Not sure about how commits are split but looks good :)

PR #73, opened 2022-10-01
ARM: Implement `CMP` ALU Instruction

fedebuonco:
Straightforward implementation of the CMP instruction. Based on the previous implementation of TEQ. More test should be written.

PS: This is the second PR on this topic, I made a mistake and closed the previous PR #68


guerinoni (commented):
Looks good but it needs rebase and also you can add the case about S=0 :D

PR #80, opened 2022-10-11
Refactor shift operand

AlessioC31:
I've refactored a bit and completed the shift op2 logic.

* Added `Type` trait for ALU instructions to know if it is logical or arithmetical
* I moved the `get_operand` logic to a separate function
* I refactored a bit the `shift` logic: now barrel shifter operations correctly set the carry flag when the instruction is logical and `S` is set

I know that `09a20b5b9f95fbc6f301b13b321c79b194b1f9f7` is big but it doesn't make sense to me to split it, since it's all about the same task, but let me know :)

I think the barrel shifter logic is complex enough to be worth quite some tests. I didn't have time yet to add them but they're for sure needed.

### DIFF
--- a/emu/src/alu_instruction.rs
+++ b/emu/src/alu_instruction.rs
@@ -17,6 +17,26 @@ pub enum ArmModeAluInstruction {
     Mvn = 0xF,
 }
 
+#[derive(Eq, PartialEq, Debug)]
+pub enum AluInstructionKind {
+    Logical,
+    Arithmetic,
+}
+
+pub trait Kind {
+    fn kind(&self) -> AluInstructionKind;
+}
+
+impl Kind for ArmModeAluInstruction {
+    fn kind(&self) -> AluInstructionKind {
+        use ArmModeAluInstruction::*;
+        match &self {
+            And | Eor | Tst | Teq | Orr | Mov | Bic | Mvn => AluInstructionKind::Logical,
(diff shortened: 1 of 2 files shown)


emu/src/alu_instruction.rs
```diff
@@ -17,6 +17,26 @@ pub enum ArmModeAluInstruction {
...
+    Logical,
+    Arithmetic,
+}
+
+pub trait Type {
+    fn get_type(&self) -> AluInstructionTypes;
+}
+
+impl Type for ArmModeAluInstruction {
+    fn get_type(&self) -> AluInstructionTypes {
```
guerinoni, line 31: Better
```suggestion
    fn type(&self) -> AluInstructionTypes {
```
or
```suggestion
    fn kind(&self) -> AluInstructionTypes {
```

guerinoni (commented):
Seems good!
Little suggestion, in the commit messages you missed capital letter and turn in present simple, in the body you can se the past.
Another suggestion in "use meaningful message" like "Split `shift` as separated function, instead of generic "refactor".

Last thing, missing test for "kind"

emu/src/alu_instruction.rs
AlessioC31: I like `type` more but it seems like you can't have a function called `type` in rust.
I don't like `kind` that much since I think the term type is more appropriate in this case. What do you think?

guerinoni: Ok for me :)

guerinoni (approved):
LGTM

PR #81, opened 2022-10-16
ARM: Implement Block Data Transfer (LDM)

guerinoni:
This is the first implementation of LDM for a specific case encountered,
we need to improve our test suite for covering more cases.

Co-authored-by: Alessio Cosenza <cosenzaalessio17@gmail.com>

### DIFF
--- a/emu/src/opcode.rs
+++ b/emu/src/opcode.rs
@@ -54,6 +54,9 @@ impl Display for ArmModeOpcode {
             ArmModeInstruction::Branch | ArmModeInstruction::BranchLink => {
                 "FMT: |_Cond__|1_0_1|L|___________________Offset______________________|\n"
             }
+            ArmModeInstruction::BlockDataTransfer => {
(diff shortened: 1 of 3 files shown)


emu/src/opcode.rs
```diff
@@ -122,6 +125,9 @@ impl Display for ArmModeOpcode {
                      BIN: |{cond:07b}|1_0_1|{l:01b}|{offset:047b}|\n"
                 )
             }
+            ArmModeInstruction::BlockDataTransfer => {
+                format!("just skip this for now")
+            }
```
AlessioC31, line 130: Would it be useful to implement it now?

guerinoni: This part of code is disabled right now because it creates a little confusing to me, then right now is enough to keep raw data

AlessioC31: Got it. I didn’t notice it was disabled.

AlessioC31 (approved):

PR #83, opened 2022-10-18
Fix `PC` usage in ALU ops and fix one shift register case

AlessioC31:
This mainly fixes the behavior when using `PC` as an operand in ALU ops.

From documentation:

> The PC value will be the address of the instruction, plus 8 or 12 bytes due to instruction prefetching. If the shift amount is specified in the instruction, the PC will be 8 bytes ahead. If a register is used to specify the shift amount the PC will be 12 bytes ahead

I'm not 100% sure whether the +12 offset has to be used when `PC` is used as the first operand (no shift involved with it). But since the documentation doesn't specify I assume the same logic applies to both the first and second operands. This is confirmed by GBATEK.

This PR also adds a check when the value read from the shift register is 0. In this case, the documentation specifies that the value in `Rm` has to be used as is, without modification.

### DIFF
--- a/emu/src/arm7tdmi.rs
+++ b/emu/src/arm7tdmi.rs
@@ -148,9 +148,15 @@ impl Arm7tdmi {
             false => {
                 // bits [0-3] 2nd Operand Register (R0..R15) (including PC=R15)
                 let rm = op2.get_bits(0..=3);
-                let rm = self.registers.register_at(rm.try_into().unwrap());
                 // bit [4] - is Shift by Register Flag (0=Immediate, 1=Register)
                 let r = op2.get_bit(4);
+                let offset = match rm {
+                    // if Rm is R15(PC) we need to offset its value because of
+                    // instruction pipelining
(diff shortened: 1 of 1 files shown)


emu/src/arm7tdmi.rs
```diff
@@ -168,7 +168,14 @@ impl Arm7tdmi {
...
 
-                        self.registers.register_at(rs.try_into().unwrap()) & 0xFF
+                        let rs = self.registers.register_at(rs.try_into().unwrap()) & 0xFF;
+
+                        // If shift is taken from register and the value is 0 Rm is directly used as operand
+                        if rs == 0 {
+                            return rm;
+                        }
+
+                        rs
```
guerinoni, line 178: Can we write a little test about this?

AlessioC31: Yup, forgot about it. I've added it to the latest commit. Thanks

guerinoni (approved):
LGTM

PR #91, opened 2022-10-26
ALU: Add `SUB`, `RSB`, `CMN` op and Fix `CMP`, `MOV` op

AlessioC31:
* Add support for `SUB`, `RSB` and `CMN` ALU op
* Fix the `CMP` op: it was not executing all the sub logic (overflow/carry check, etc.)
* Fix `MOV` op: it was not setting the `CPSR` flags
* Extracted the add and sub logic to two separate functions since they are used in many operations (`SUB` and `CMP`, `ADD` and `CMN`) and there will be more with the same logic: `ADC`, `SBC`, `RSC`. I created a new struct to return the result of an arithmetic operation.

### DIFF
--- a/emu/src/arm7tdmi.rs
+++ b/emu/src/arm7tdmi.rs
@@ -267,6 +267,7 @@ impl Arm7tdmi {
             ArmModeAluInstruction::Add => self.add(rd.try_into().unwrap(), rn, op2, s),
             ArmModeAluInstruction::Orr => self.orr(rd.try_into().unwrap(), rn, op2, s),
             ArmModeAluInstruction::Sub => self.sub(rd.try_into().unwrap(), rn, op2, s),
+            ArmModeAluInstruction::Rsb => self.rsb(rd.try_into().unwrap(), rn, op2, s),
             _ => todo!("implement alu operation: {}", alu_op_code),
         }
     }
@@ -506,6 +507,10 @@ impl Arm7tdmi {
         }
     }
 
+    fn rsb(&mut self, rd: usize, rn: u32, op2: u32, s: bool) {
+        self.sub(rd, op2, rn, s);
+    }
+    
     fn shift_operand(
         &mut self,
         alu_opcode: u32,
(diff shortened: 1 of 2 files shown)


emu/src/arm7tdmi.rs
```diff
@@ -443,8 +443,13 @@ impl Arm7tdmi {
...
     }
 
-    fn mov(&mut self, rd: usize, op2: u32) {
+    fn mov(&mut self, rd: usize, op2: u32, s: bool) {
         self.registers.set_register_at(rd, op2);
+
+        if s {
+            self.cpsr.set_zero_flag(op2 == 0);
+            self.cpsr.set_sign_flag(op2.get_bit(31));
+        }
```
guerinoni, line 477: Test? :)

emu/src/arm7tdmi.rs
```diff
@@ -460,10 +490,57 @@ impl Arm7tdmi {
...
+        let sub_result = Self::sub_inner_op(rn, op2);
+
+        self.registers.set_register_at(rd, sub_result.result);
+
+        if s {
+            self.cpsr.set_overflow_flag(sub_result.overflow);
+            self.cpsr.set_carry_flag(sub_result.carry);
+            self.cpsr.set_zero_flag(sub_result.zero);
+            self.cpsr.set_sign_flag(sub_result.sign);
+        }
```
guerinoni, line 539: Would love to see `sub_result` passed directly to `cpsr` in some way, because seems an over engineering with this extra struct. Using a separated fn for creating a struct and other bunch of line for assigning... Don't sure we have a better design

AlessioC31: The separate fn is in my opinion needed since the add/sub logic is used in many ops. What would you return from this fn? At first, I thought about a tuple but a struct seemed more clear since it has named properties.

We could add a fn in `Cpsr` that accepts an `ArithmeticOpResult` and sets the flags?

I don't know what would be the best solution in this case

guerinoni: We can go for something like `self.cpsr.apply_flags(flags_result)`

AlessioC31: Done, thanks!

guerinoni (approved):

PR #92, opened 2022-10-26
`Timer registers` implementation for `Internal memory`

FedericoBruzzone:
[GBA-Timers](https://rust-console.github.io/gbatek-gbaonly/#--gba-timers)


FedericoBruzzone:
image
These parts will need to be implemented

emu/src/timer_registers.rs
```diff
@@ -0,0 +1,132 @@
+use crate::bitwise::Bits;
+
+pub enum TimerRegisterAccessControl {
+    Read,
+    Write,
+    ReadWrite,
+}
```
AlessioC31, line 7: I think it would be better to define an enum like `IoRegisterAccessControl` and use it both here and in `lcd_registers` and future IO registers, instead of creating new enums every time.

emu/src/timer_registers.rs
```diff
@@ -0,0 +1,132 @@
...
+            }
+
+            _ => panic!("Requested Timer Register cannot be written."),
+        }
+    }
+
+    pub const fn with_access_control(access: TimerRegisterAccessControl) -> Self {
+        Self { value: 0, access }
+    }
+}
```
AlessioC31, line 45: I think there's no need to define this struct. In file `io_registers.rs` there's `IORegister` struct

emu/src/internal_memory.rs
```diff
@@ -164,6 +197,7 @@ impl IoDevice for InternalMemory {
         match address {
             0x03000000..=0x03007FFF => self.internal_work_ram[(address - 0x03000000) as usize],
             0x04000000..=0x04000055 => self.read_address_lcd_register(address),
+            0x04000100..=0x0400010E => self.read_address_timer_register(address), // @ FB
```
AlessioC31, line 200: Timer registers memory goes up to `0x0400010F` since at `0x0400010E` there's `TM3CNT_H` which is 2 bytes long

emu/src/internal_memory.rs
```diff
@@ -177,6 +211,7 @@ impl IoDevice for InternalMemory {
                 self.internal_work_ram[(address - 0x03000000) as usize] = value
             }
             0x04000000..=0x04000055 => self.write_address_lcd_register(address, value),
+            0x04000100..=0x0400010E => self.write_address_timer_register(address, value), // @ FB
```
AlessioC31, line 214: Same as above

emu/src/internal_memory.rs
```diff
@@ -154,6 +159,34 @@ impl InternalMemory {
...
+    fn read_address_timer_register(&self, address: u32) -> u8 {
+        match address {
+            0x04000100 => self.timer_registers.tm0cnt_l.read().get_byte(0),
+            0x04000102 => self.timer_registers.tm0cnt_h.read().get_byte(0),
+            0x04000104 => self.timer_registers.tm1cnt_l.read().get_byte(0),
+            0x04000106 => self.timer_registers.tm1cnt_h.read().get_byte(0),
+            0x04000108 => self.timer_registers.tm2cnt_l.read().get_byte(0),
+            0x0400010A => self.timer_registers.tm2cnt_h.read().get_byte(0),
+            0x0400010C => self.timer_registers.tm3cnt_l.read().get_byte(0),
+            0x0400010E => self.timer_registers.tm3cnt_h.read().get_byte(0),
```
AlessioC31, line 186: Why do we never read the higher part of the registers?

I would expect something like:

```rust
0x04000100 => self.timer_registers.tm0cnt_l.set_byte(0, value),
0x04000101 => self.timer_registers.tm0cnt_l.set_byte(1, value),
0x04000102 => self.timer_registers.tm0cnt_l.set_byte(0, value),
...
```

Maybe there's a reason why we never access these addresses?

AlessioC31 (changes requested):
Would it be possible to add a `TODO` or a panic somewhere in order to remember that we still need to implement the part you highlighted in the previous comment?

emu/src/internal_memory.rs
AlessioC31: Ok I thing I got it why, it says that bits from 8 to 15 are not used, correct?

emu/src/timer_registers.rs
FedericoBruzzone: Perfect!

emu/src/timer_registers.rs
FedericoBruzzone: Perfect!

AlessioC31:
I think we can squash the 2 commits into 1, what do you think?

emu/src/internal_memory.rs
FedericoBruzzone: image

Correct me if I'm wrong, but the bit from `8..-15` is not used.

AlessioC31: Yep, that's what I meant in the previous comment. I can't "resolve" this (and the 2 previous ones) comment, I think only you can?

FedericoBruzzone: Yes, I think It works like this! :D

FedericoBruzzone: What would you like to do? Resolve what?

FedericoBruzzone: `Timer registers memory goes up to 0x0400010F since at 0x0400010E there's TM3CNT_H which is 2 bytes long` this and `Same as above`?

AlessioC31: Yes, sorry for the confusion :)

FedericoBruzzone:
Maybe it is not necessary, but let us talk about it :D

FedericoBruzzone:
Before touching my code, it would be a good idea to separate `IoRegister` from `LCDRegisters`?

emu/src/internal_memory.rs
FedericoBruzzone: :+1:

emu/src/internal_memory.rs
FedericoBruzzone: No problem mate! :+1:

AlessioC31:
> Before touching my code, it would be a good idea to separate `IoRegister` from `LCDRegisters`?

You mean to have `IORegister` and `IORegisterAccessControl` in a separate file and then have specific files for `LCDRegisters`, `TimerRegisters` etc? I think it would be a good idea. But can we group them under the same directory to avoid having too many files under the same directory? I think it would help navigate the code

FedericoBruzzone:
OMG, you're right, how is it possibile? 
I'll delete the second commit

FedericoBruzzone:
I mean that it would be a good idea create a new directory called `registers` within: `IoRegister`, `LCDRegister`, `TimerRegister` and so on.

AlessioC31:
> I mean that it would be a good idea create a new directory called `registers` within: `IoRegister`, `LCDRegister`, `TimerRegister` and so on.

Yes, I think it would be good. Maybe I would not call it `registers` since the term register can also refer to cpu registers. Something like `io_registers`? But then how would we call the file that contains `IORegister` and `IORegisterAccessControl`? 🤔

FedericoBruzzone:
Perfect, I'll fix everything tonight. 
A good name for the directory could be `external_register`, in this way we don't need to change the name of `io_registers`. What do you think about it?

emu/src/timer_registers.rs
guerinoni: Agreed

emu/src/internal_memory.rs
guerinoni: I would love to see some comment in code for future doubt :)

emu/src/internal_memory.rs
guerinoni, line 200: What is `@ FB`?

guerinoni (changes requested):

emu/src/internal_memory.rs
FedericoBruzzone: It was my comment, I'll delete it!

emu/src/internal_memory.rs
FedericoBruzzone: It's not a problem!

emu/src/io_register_internal_memory/timer_registers.rs
```diff
@@ -0,0 +1,88 @@
...
+
+        reg.write(5);
+
+        assert_eq!(reg.read(), 5);
+
+        reg.set_byte(1, 3);
+
+        assert_eq!(reg.read(), (3 << 8) | 5);
+    }
+}
```
AlessioC31, line 88: We don't need these tests right? They are already in `emu/src/io_register_internal_memory/io_registers.rs`

AlessioC31 (changes requested):

AlessioC31 (approved):

AlessioC31 (approved):

guerinoni (approved):

PR #93, opened 2022-10-27
Memory: Add bios memory

guerinoni:

### DIFF
--- a/emu/src/internal_memory.rs
+++ b/emu/src/internal_memory.rs
@@ -3,6 +3,9 @@ use crate::io_device::IoDevice;
 use crate::io_registers::LCDRegisters;
 
 pub struct InternalMemory {
+    /// From 0x00000000 to 0x00003FFF (16 KBytes)
+    bios_system_rom: [u8; 0x4000],
+
     /// From 0x03000000 to 0x03007FFF (32kb).
     internal_work_ram: [u8; 0x8000],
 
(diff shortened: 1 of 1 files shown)


FedericoBruzzone:
Looks good. 
Let's talk about if it's better to merge #92 before this or before this one.

AlessioC31 (approved):
Looks good to me too.

Out of curiosity, do you know who should populate this memory region? Because I've read that GBA stores in BIOS some procedures (lz compression, division, etc). Should it be read from a bios file? Or all the `.gba` files populate it?

FedericoBruzzone (approved):

emu/src/internal_memory.rs
```diff
@@ -3,6 +3,9 @@ use crate::io_device::IoDevice;
 use crate::io_registers::LCDRegisters;
 
 pub struct InternalMemory {
+    /// From 0x00000000 to 0x00003FFF (16 KBytes)
+    bios_system_rom: [u8; 0x3FFF],
```
gallottino, line 7: The size is wrong. The right one is `0x4000`. I open a issue about this #97.

emu/src/internal_memory.rs
```diff
@@ -28,6 +31,7 @@ impl Default for InternalMemory {
 impl InternalMemory {
     pub const fn new() -> Self {
         Self {
+            bios_system_rom: [0; 0x3FFF],
```
gallottino, line 34: Same of above.

gallottino (changes requested):

emu/src/internal_memory.rs
guerinoni: Done :)

emu/src/internal_memory.rs
guerinoni: Done :)

gallottino (approved):

AlessioC31 (approved):

PR #98, opened 2022-10-28
Fix memory buffers in `InternalMemory`

### LINKED ISSUE #97: Wrong size of buffer memory in `internal_memory.rs`
It the documentation  we can find:

> 03000000-03007FFF   WRAM - On-chip Work RAM   (32 KBytes)
> 05000000-050003FF   BG/OBJ Palette RAM        (1 Kbyte)
> 06000000-06017FFF   VRAM - Video RAM          (96 KBytes)

but we define them as an array of (take WRAM as example): `[u8; 0x7FFF]`

0x7FFF =  32767 bytes, but we are missing 1 bytes to get 32 Kbytes!
The right value in this case is **0x8000**

This is a problem since we forget to add the last byte in the array. The second parameter is the size of the array (in this case 32 Kbytes) and it is not the last index used to access. 

We need to change all the arrays in `InternalMemory`.

p.s. preparatory to #89

gallottino:
Closes #97

### DIFF
--- a/emu/src/internal_memory.rs
+++ b/emu/src/internal_memory.rs
@@ -4,16 +4,16 @@ use crate::io_registers::LCDRegisters;
 
 pub struct InternalMemory {
     /// From 0x03000000 to 0x03007FFF (32kb).
-    internal_work_ram: [u8; 0x7FFF],
+    internal_work_ram: [u8; 0x8000],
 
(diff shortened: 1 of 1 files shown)


guerinoni (approved):
Looks good, not sure why we used 1 bit less -_-

AlessioC31 (approved):
Looks good to me too 😅

PR #99, opened 2022-10-29
Palette visualizer

### LINKED ISSUE #89: Palette Visualizer
I checked in the documentation and GBA has many `MODES` to render the screen.
I'm starting to study the `MODE 0` and `MODE 1`: they use palettes to render some tiles built by memory indices referring to palette RAM.

It can be useful to have a widget to visualize this part of memory and to check colors and palette(s) inside. This is also preparatory to implement the real render algorithm in `ppu.rs`

From the documentation:
> **Color Palette RAM**
> Each BG and OBJ palette RAM may be either split into 16 palettes with 16 colors each, or may be used as a single palette with 256 colors. 
> 
> **Color Definitions**
> Each color occupies two bytes (same for 327688 BG modes):
> [0-4]         Red Intensity    (0-31)
> [5-9]         Green Intensity (0-31)
> [10-14]     Blue Intensity    (0-31)
> 15          Not used

gallottino:
Closes #89 

I also added a feature  _"debug"_ to put random values in the `palette_ram`. In can be useful to use this features also to some hard coded test for the render process.

To run clementine in debug mode you just need to run the command:
`just debug *rom_path/rom.gba*`

### DIFF
--- a/emu/src/ppu.rs
+++ b/emu/src/ppu.rs
@@ -1,2 +1,122 @@
+use std::{cell::RefCell, rc::Rc};
+
+#[cfg(feature = "debug")]
+use rand::Rng;
+
 pub const DISPLAY_WIDTH: usize = 240;
 pub const DISPLAY_HEIGHT: usize = 160;
+pub const PALETTE_SIZE: usize = 16;
+
+pub const BG_PALETTE_ADDRESS: i32 = 0x05000000;
+pub const OBJ_PALETTE_ADDRESS: i32 = 0x05000200;
+
+use crate::{
+    memory::internal_memory::InternalMemory,
+    render::palette_color::{
+        PaletteColor, PaletteType, MAX_COLORS_FULL_PALETTE, MAX_COLORS_SINGLE_PALETTE,
+        MAX_PALETTES_BY_TYPE,
+    },
+};
+
+pub struct PixelProcessUnit {
+    internal_memory: Rc<RefCell<InternalMemory>>,
+}
+
+impl PixelProcessUnit {
+    pub fn new(internal_memory: Rc<RefCell<InternalMemory>>) -> Self {
+        Self { internal_memory }
+    }
+
+    fn get_array_color(&self, index: usize, palette_type: &PaletteType) -> [u8; 2] {
+        debug_assert!(index % 2 == 0);
+
+        match palette_type {
+            PaletteType::BG => [
+                self.internal_memory.borrow().bg_palette_ram[index],
+                self.internal_memory.borrow().bg_palette_ram[index + 1],
+            ],
+            PaletteType::OBJ => [
+                self.internal_memory.borrow().obj_palette_ram[index],
+                self.internal_memory.borrow().obj_palette_ram[index + 1],
+            ],
+        }
+    }
+
+    fn get_color_from_full_palette(
+        &self,
+        color_index: usize,
+        palette_type: &PaletteType,
+    ) -> PaletteColor {
+        debug_assert!(color_index < MAX_COLORS_FULL_PALETTE);
+        self.get_array_color(color_index * 2, palette_type).into()
+    }
+
+    fn get_color_from_single_palette(
+        &self,
+        palette_index: usize,
+        color_index: usize,
+        palette_type: &PaletteType,
+    ) -> PaletteColor {
+        debug_assert!(palette_index < MAX_PALETTES_BY_TYPE);
+        debug_assert!(color_index < MAX_COLORS_SINGLE_PALETTE);
+
+        self.get_color_from_full_palette(
+            (palette_index * MAX_COLORS_SINGLE_PALETTE) + color_index,
+            palette_type,
+        )
+    }
+
+    pub fn get_palettes(&self, palette_type: &PaletteType) -> Vec<Vec<PaletteColor>> {
+        let mut palettes = vec![];
+
+        for palette_index in 0..MAX_PALETTES_BY_TYPE {
+            palettes.push(vec![]);
+            for color_index in 0..MAX_COLORS_SINGLE_PALETTE {
+                palettes[palette_index as usize].push(self.get_color_from_single_palette(
+                    palette_index,
+                    color_index,
+                    palette_type,
+                ));
+            }
+        }
+        palettes
+    }
+
+    #[cfg(feature = "debug")]
+    pub fn load_random_palettes(&mut self) {
+        for i in 0..200 {
+            let mut rng = rand::thread_rng();
+            let mut value: u8 = rng.gen();
+            self.internal_memory.borrow_mut().bg_palette_ram[i] = value;
+            value = rng.gen();
+            self.internal_memory.borrow_mut().obj_palette_ram[i] = value;
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::PixelProcessUnit;
+    use crate::{memory::internal_memory::InternalMemory, render::palette_color::PaletteType};
+    use std::{cell::RefCell, rc::Rc};
+
+    #[test]
+    fn test_get_palettes_bg() {
+        let internal_memory = Rc::new(RefCell::new(InternalMemory::new()));
+        let ppu = PixelProcessUnit::new(internal_memory);
+
+        let bg_palettes = ppu.get_palettes(&PaletteType::BG);
+        assert_eq!(bg_palettes.len(), 16);
+        assert_eq!(bg_palettes[0].len(), 16);
+    }
+
+    #[test]
+    fn test_get_palettes_obj() {
+        let internal_memory = Rc::new(RefCell::new(InternalMemory::new()));
+        let ppu = PixelProcessUnit::new(internal_memory);
+
+        let obj_palettes = ppu.get_palettes(&PaletteType::OBJ);
+        assert_eq!(obj_palettes.len(), 16);
+        assert_eq!(obj_palettes[0].len(), 16);
+    }
+}
--- a/emu/src/render/palette_color.rs
+++ b/emu/src/render/palette_color.rs
@@ -1,0 +1,143 @@
+use std::fmt::Display;
+
+use crate::bitwise::Bits;
+
+/// Each palette can contains 16 colors
+pub const MAX_COLORS_SINGLE_PALETTE: usize = 16;
+
+/// BG palettes and OBJ palettes can be use as a single palette
+pub const MAX_COLORS_FULL_PALETTE: usize = 256;
+
+/// Number of max palettes both for BG and OBG
+pub const MAX_PALETTES_BY_TYPE: usize = 16;
+
+/// Memory info about BG palette
+pub const BG_PALETTE_ADDRESS: u32 = 0x05000000;
+pub const _BG_PALETTE_SIZE: usize = 0x200;
+
+/// Memory info about OBJ palette
+pub const OBJ_PALETTE_ADDRESS: u32 = 0x05000200;
+pub const _OBJ_PALETTE_SIZE: usize = 0x200;
+
+#[derive(PartialEq, Eq, Clone)]
+pub enum PaletteType {
+    BG,
+    OBJ,
+}
+
+pub struct PaletteColor {
+    pub red: u8,
+    pub green: u8,
+    pub blue: u8,
+}
+
+impl Display for PaletteColor {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "({},{},{})", self.red, self.green, self.blue)
+    }
+}
+
+impl From<u16> for PaletteColor {
+    fn from(color: u16) -> Self {
+        // Color     Values     Bytes
+        //-------------------------------
+        // red ---> [0 - 31]    0-4 bit
+        // green -> [0 - 31]    5-9 bit
+        // blue --> [0 - 31]    10-14 bit
+        // useless -------->    15 bit
+        // bits are taken from upperbyte + lowerbyte u16
+
+        let red = color.get_bits(0..=4);
+        let green = color.get_bits(5..=9);
+        let blue = color.get_bits(10..=14);
+
+        Self {
+            red: red as u8,
+            green: green as u8,
+            blue: blue as u8,
+        }
+    }
+}
+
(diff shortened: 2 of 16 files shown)


emu/src/ppu.rs
```diff
@@ -1,2 +1,133 @@
...
+    pub fn get_bg_palette_ram(&self) -> [u8; 0x200] {
+        self.internal_memory.borrow_mut().bg_palette_ram
+    }
+
+    pub fn get_obj_palette_ram(&self) -> [u8; 0x200] {
+        self.internal_memory.borrow_mut().obj_palette_ram
+    }
+
+    pub fn color_from_two_u8(&self, upper_byte: u8, lower_byte: u8) -> PaletteColor {
+        let mut upper: u32 = upper_byte.into();
```
gallottino, line 51: I have some doubts about this: is it better to implement `Bits` trait also for u8?

gallottino:
Work in progress stuff:
- [ ] add documentation in `ppu.rs`
- [ ] add tests in `ppu.rs`

emu/src/ppu.rs
```diff
@@ -1,2 +1,133 @@
 pub const DISPLAY_WIDTH: usize = 240;
 pub const DISPLAY_HEIGHT: usize = 160;
+pub const PALETTE_SIZE: usize = 16;
+
+pub const BG_PALETTE_ADDRESS: u32 = 0x05000000;
+pub const OBJ_PALETTE_ADDRESS: u32 = 0x05000200;
```
guerinoni, line 6: Maybe better declare those after `use...`

emu/src/ppu.rs
```diff
@@ -1,2 +1,133 @@
...
+    render::palette_color::{PaletteColor, PaletteType},
+};
+
+pub struct PixelProcessUnit {
+    internal_memory: Rc<RefCell<InternalMemory>>,
+}
+
+impl PixelProcessUnit {
+    pub fn new(internal_memory: Rc<RefCell<InternalMemory>>) -> Self {
+        Self { internal_memory }
```
guerinoni, line 24: In other iteration I would like to see this referring only to the part of video and not to entire `InternalMemory`, not important now, maybe we can just spit "logical memory" from "video memory"

ui/src/palette_visualizer.rs
```diff
@@ -0,0 +1,132 @@
...
+use emu::{
+    ppu::{BG_PALETTE_ADDRESS, OBJ_PALETTE_ADDRESS},
+    render::palette_color::{PaletteColor, PaletteType},
+};
+use std::sync::{Arc, Mutex};
+
+use emu::{arm7tdmi::Arm7tdmi, gba::Gba};
+
+pub struct PaletteVisualizer {
+    gba: Arc<Mutex<Gba<Arm7tdmi>>>,
```
guerinoni, line 13: Maybe this should be `Arc<Mutex<Gba<Cpu>>>` or similar

ui/src/palette_visualizer.rs
```diff
@@ -0,0 +1,132 @@
...
+                use View as _;
+                self.ui(ui);
+            });
+    }
+}
+
+impl View for PaletteVisualizer {
+    fn ui(&mut self, ui: &mut egui::Ui) {
+        if let Ok(gba) = self.gba.lock() {
+            self.palettes = gba.cpu.ppu.get_palettes(self.palette_type.clone());
```
guerinoni, line 50: I think we can pass `&self.palette_type`

guerinoni (commented):
First round of review... There are some commits with "more" things in same context but... :P

emu/src/render/palette_color.rs
```diff
@@ -0,0 +1,128 @@
...
+
+impl Display for PaletteColor {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "({},{},{})", self.red, self.green, self.blue)
+    }
+}
+
+impl From<u16> for PaletteColor {
+    fn from(color_u16: u16) -> Self {
+        let color_u32: u32 = color_u16.into();
```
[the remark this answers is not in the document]
gallottino: Not sure too. The problem is that u32 is not castable to PaletteColor as the color need to have only 15 bits and also we implemented the `Bits` trait only for u32.

emu/src/ppu.rs
gallottino: I agree :)

ui/src/palette_visualizer.rs
gallottino: Yes, but maybe `PaletteVisualizer` is a only Arm7tdmi tool. In other case we need to put the `InternalMemory` outside cpu or we need to implement in the trait a method for get it.

FedericoBruzzone (approved):

FedericoBruzzone (approved):

FedericoBruzzone (approved):

emu/src/render/palette_color.rs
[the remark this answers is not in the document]
AlessioC31: Can we implement `Bits` for `u16` too? I think it makes sense.

emu/src/render/palette_color.rs
```diff
@@ -0,0 +1,129 @@
+use std::fmt::Display;
+
+use crate::bitwise::Bits;
+
+#[derive(PartialEq, Eq, Clone)]
+pub enum PaletteType {
+    BG,
+    OBJ,
+    FULL,
```
AlessioC31, line 9: What to we use `FULL` for?

emu/src/ppu.rs
```diff
@@ -1,2 +1,130 @@
...
+                self.internal_memory.borrow().bg_palette_ram[index],
+                self.internal_memory.borrow().bg_palette_ram[index + 1],
+            ],
+            PaletteType::OBJ => [
+                self.internal_memory.borrow().obj_palette_ram[index],
+                self.internal_memory.borrow().obj_palette_ram[index + 1],
+            ],
+            PaletteType::FULL => match index {
+                0..=255 => self.get_array_color(index * 2, PaletteType::BG),
+                256..=551 => self.get_array_color((index - 256) * 2, PaletteType::OBJ),
```
AlessioC31, line 39: Why `551`? We have 512 colors in total, right?

emu/src/render/palette_color.rs
AlessioC31: Ok, I think it is because the documentation says:

> Each BG and OBJ palette RAM may be either split into 16 palettes with 16 colors each, or may be used as a single palette with 256 colors.

So `FULL` is to access both areas as a single palette with 512 colors right? But wouldn't it be better to split this behavior into `BG_FULL` and `OBJ_FULL`?

Correct me if I'm wrong! Thanks!

gallottino: Ah! yeah you are right, I missunderstood the docs.

emu/src/ppu.rs
gallottino: Yes, my bad :)

gallottino:
> Can we implement Bits for u16 too? I think it makes sense. 

@AlessioC31 
I don't know why I can't answer to your comment about implementing Bits traits for u16.

Anyway I also was thinking about it but we have the same methods  both for u32 and u16 and it seems to me a bit redundant. We can think a solution, but I prefer open an issue for this and iterate later. But I totally agree with you that it is a weird cast.

guerinoni:
> > Can we implement Bits for u16 too? I think it makes sense.
> 
> @AlessioC31 I don't know why I can't answer to your comment about implementing Bits traits for u16.
> 
> Anyway I also was thinking about it but we have the same methods both for u32 and u16 and it seems to me a bit redundant. We can think a solution, but I prefer open an issue for this and iterate later. But I totally agree with you that it is a weird cast.
> 
> P.s. Just a flash thought: we can force that the type for the `Bits` trait must implement `From<u16>` or something like this. But I would like to don't repeat the same code for all the u* types.

You can implement both and from u32 implementation call u16 function, in this way the core of logic it will stay only in one place :)

gallottino:
@guerinoni @AlessioC31 
I fixed some stuff. 

1. Now `Bits` trait can be used with all the uint types. I did a generical abstraction for all the cases.
2. I removed `FULL`  palette type at all
3. `From<u16> for PaletteColor` not doesn't cast anymore to u32
4. Minor stuff written in the previous comments

AlessioC31 (commented):
Why do we pass &PaletteType instead of PaletteType in some functions?

gallottino:
> Why do we pass &PaletteType instead of PaletteType in some functions?

To avoid `clone()` when you pass it many times in other functions. In other case, without clone the variable exits from the scope and it be lost. With `&PaletteType` we don't copy uselessly the value, but we are just passing its reference. For this specific case the difference is not so big because we don't need to pass by value a big amount of data, but anyway I think, when it is possible, this approach is better.

AlessioC31:
> > Why do we pass &PaletteType instead of PaletteType in some functions?
> 
> To avoid `clone()` when you pass it many times in other functions. In other case, without clone the variable exits from the scope and it will be lost. With `&PaletteType` we don't copy uselessly the value, but we are just passing its reference. For this specific case the difference is not so big because we don't need to pass by value a big amount of data, but anyway I think, when it is possible, this approach is better.

But in this case `PaletteType` is practically an `int` (or something similar I suppose). I don't know how it works underneath in rust but if copying a `PaletteType` means copying an int and passing it by ref means passing a pointer, then passing it by value would be faster (you don't have the dereferencing part and since it's an `int` it will probably be passed via registers).

EDIT:
I checked with godbolt and it seems that in rust the code for the two ways are basically the same (with opt-level=0) because even when you pass by value it still dereferences (so it doesn't work as in C :D).

Also, in the case of optimizations they compile to the same assembly.

So I don't know, I would go for a `PaletteType` since it's "cleaner" but boh, no strong opinions since it's basically the same.

gallottino:
> > > Why do we pass &PaletteType instead of PaletteType in some functions?
> > 
> > 
> > To avoid `clone()` when you pass it many times in other functions. In other case, without clone the variable exits from the scope and it will be lost. With `&PaletteType` we don't copy uselessly the value, but we are just passing its reference. For this specific case the difference is not so big because we don't need to pass by value a big amount of data, but anyway I think, when it is possible, this approach is better.
> 
> But in this case `PaletteType` is practically an `int` (or something similar I suppose). I don't know how it works underneath in rust but if copying a `PaletteType` means copying an int and passing it by ref means passing a pointer, then passing it by value would be faster (you don't have the dereferencing part and since it's an `int` it will probably be passed via registers).
> 
> EDIT: I checked with godbolt and it seems that in rust the code for the two ways are basically the same (+ a copy in the `PaletteType` case) (with opt-level=0) because even when you pass by value it still dereferences since the Enum is stored in the heap in any case.
> 
> Also, in the case of optimizations they compile to the same assembly.
> 
> So I don't know, I would go for a `PaletteType` since it's "cleaner" but boh, no strong opinions since it's basically the same.
> 

Totally agree with you that in this case there is no difference. At first I coded it with `PaletteType` for the same reason. @guerinoni ask me to change it. Maybe because it is cleaner (I don't know how much) the sign of the method, but it is less cleaner the use of it as you need to add `.clone()` at the end :) 
Personally I prefer always reference if it possible, because you don't change the ownership of the variable, you don't need to copy it and so you can use it more times.

AlessioC31:
> > > > Why do we pass &PaletteType instead of PaletteType in some functions?
> > > 
> > > 
> > > To avoid `clone()` when you pass it many times in other functions. In other case, without clone the variable exits from the scope and it will be lost. With `&PaletteType` we don't copy uselessly the value, but we are just passing its reference. For this specific case the difference is not so big because we don't need to pass by value a big amount of data, but anyway I think, when it is possible, this approach is better.
> > 
> > 
> > But in this case `PaletteType` is practically an `int` (or something similar I suppose). I don't know how it works underneath in rust but if copying a `PaletteType` means copying an int and passing it by ref means passing a pointer, then passing it by value would be faster (you don't have the dereferencing part and since it's an `int` it will probably be passed via registers).
> > EDIT: I checked with godbolt and it seems that in rust the code for the two ways are basically the same (+ a copy in the `PaletteType` case) (with opt-level=0) because even when you pass by value it still dereferences since the Enum is stored in the heap in any case.
> > Also, in the case of optimizations they compile to the same assembly.
> > So I don't know, I would go for a `PaletteType` since it's "cleaner" but boh, no strong opinions since it's basically the same.
> 
> Totally agree with you that in this case there is no difference. At first I coded it with `PaletteType` for the same reason. @guerinoni ask me to change it. Maybe because it is cleaner (I don't know how much) the sign of the method, but it is less cleaner the use of it as you need to add `.clone()` at the end :) Personally I prefer always reference if it possible, because you don't change the ownership of the variable, you don't need to copy it and so you can use it more times.

But if we derive `Copy` you don't need to use `.clone()` everytime, right? I ask because I genuinely don't know

emu/src/render/palette_color.rs
```diff
@@ -0,0 +1,143 @@
...
+
+/// BG palettes and OBJ palettes can be use as a single palette
+pub const MAX_COLORS_FULL_PALETTE: usize = 256;
+
+/// Number of max palettes both for BG and OBG
+pub const MAX_PALETTES_BY_TYPE: usize = 16;
+
+/// Memory info about BG palette
+pub const BG_PALETTE_ADDRESS: u32 = 0x05000000;
+pub const _BG_PALETTE_SIZE: usize = 0x200;
```
guerinoni, line 16: Not used?

emu/src/render/palette_color.rs
```diff
@@ -0,0 +1,143 @@
...
+
+/// Memory info about OBJ palette
+pub const OBJ_PALETTE_ADDRESS: u32 = 0x05000200;
+pub const _OBJ_PALETTE_SIZE: usize = 0x200;
```
guerinoni, line 20: Also this, is it not used?

guerinoni (approved):
Don't know if we miss some tests for `Bits` trait on all cases (u8, 816, u32, u128)

emu/src/render/palette_color.rs
gallottino: Yes, I need it for later.

emu/src/render/palette_color.rs
gallottino: Same of above.

PR #101, opened 2022-10-29
Improve Single Data Transfer

guerinoni:
- ARM: Add kind of read/write for LDR/STR
- ARM: Move `single_data_transfer` in dedicated file


emu/src/single_data_transfer.rs
```diff
@@ -0,0 +1,161 @@
...
+            address.wrapping_sub(offset)
+        } else {
+            address.wrapping_add(offset)
+        };
+
+        match load_store {
+            SingleDataTransfer::Ldr => match byte_or_word {
+                ReadWriteKind::Byte => self
+                    .registers
+                    .set_register_at(rd.try_into().unwrap(), self.memory.read_at(address) as u32), // FIXME: writing a byte into u32 overwrite all register or just 0-7 bits?
```
AlessioC31, line 93: From documentation: 

> The selected byte is placed in the bottom 8 bits of the destination register and the remaining bits of the register are filled with zeros.

So it is correct as it is now

emu/src/single_data_transfer.rs
```diff
@@ -0,0 +1,161 @@
...
+        // bits [15-12] - Source/Destination Register
+        let rd = op_code.get_bits(12..=15);
+
+        let offset = if immediate {
+            todo!()
+        } else {
+            op_code.get_bits(0..=11)
+        };
+
+        let load_store: SingleDataTransfer = op_code.raw.into();
```
AlessioC31, line 81: Personally, I don't like the idea of doing `ArmModOpcode` -> `u32` -> `SingleDataTransfer`. Couldn't we just implement `From<ArmModeOpcode` for `SingleDataTransfer` instead of manually going through `u32`?

The same also applies to `ReadWriteKind`.

What do you think?

emu/src/arm7tdmi.rs
```diff
@@ -13,7 +13,7 @@ use crate::{cpsr::Cpsr, cpu::Cpu};
 /// Contains the 16 registers for the CPU, latest (R15) is special because
 /// is the program counter.
 #[derive(Default)]
-struct Registers([u32; 16]);
+pub struct Registers([u32; 16]);
```
AlessioC31, line 16: Why do we sometimes use `pub` and sometimes `pub(crate)`?

I ask because I'm a rust noob :)

guerinoni: https://doc.rust-lang.org/reference/visibility-and-privacy.html
Basically `pub` is public also for other crates (if you expose your crate)
`pub(crate)` makes public only within the same crate but for other modules

emu/src/single_data_transfer.rs
guerinoni: We can change it :)

emu/src/single_data_transfer.rs
guerinoni: Nice, I'll remove the comment

AlessioC31: I still see the comments `FIXME` comments, but maybe is github I don't know

emu/src/single_data_transfer.rs
```diff
@@ -0,0 +1,165 @@
...
+        match load_store {
+            SingleDataTransfer::Ldr => match byte_or_word {
+                ReadWriteKind::Byte => self
+                    .registers
+                    .set_register_at(rd.try_into().unwrap(), self.memory.read_at(address) as u32), // FIXME: writing a byte into u32 overwrite all register or just 0-7 bits?
+                ReadWriteKind::Word => todo!(),
+            },
+            SingleDataTransfer::Str => {
+                match byte_or_word {
+                    ReadWriteKind::Byte => self.memory.write_at(address, rd as u8), // FIXME: Is it right to truncate rd to u8?
```
AlessioC31, line 99: Documentation says

> A byte store (STRB) repeats the bottom 8 bits of the source register four times across data bus outputs 31 through 0. The external memory system should activate the appropriate byte subsystem to store the data.

But I don’t know what this means :)

AlessioC31: I think it is correct that it truncates to the first 8 bits but then I don’t know what it means that it repeats the 8 bits 4 times in the data bus.

FedericoBruzzone (approved):

emu/src/single_data_transfer.rs
FedericoBruzzone: I also read this part, and I think that the `external memory`, in the future, should activate a byte to copy this 8 bits four times to the data bus outputs. At this moment we could ignore this part, and add a comment/issue. Right?

AlessioC31: But what does it mean to copy the 8 bits four times in the data bus? Do you maybe have some documentation where it is explained how the external memory reads from the data bus?

guerinoni: I don't know what is this data bus, but if this is an option that we can activate in future, I will investigate on this

PR #104, opened 2022-11-02
Block data transfer LDR

guerinoni:
- Memory: Add unused chunk of memory
- ARM: Add LDR in block data transfer

### DIFF
--- a/emu/src/arm7tdmi.rs
+++ b/emu/src/arm7tdmi.rs
@@ -159,58 +159,37 @@ impl Arm7tdmi {
         let reg_list = op_code.get_bits(0..=15);
 
         let memory_base = self.registers.register_at(rn.try_into().unwrap());
-        let alignment = 4; // Since are word, the alignment is 4.
         let mut address = memory_base;
-        let change_address = |address: u32| {
-            if up_down {
-                address.wrapping_add(alignment)
-            } else {
-                address.wrapping_sub(alignment)
-            }
-        };
 
         if load_store {
-            for reg_destination in 0..16 {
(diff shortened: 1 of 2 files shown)


FedericoBruzzone (approved):
LGTM

emu/src/arm7tdmi.rs
```diff
@@ -152,38 +152,32 @@ impl Arm7tdmi {
...
+                arm.memory.write_at(address, value.get_bits(0..=7) as u8);
+                arm.memory
+                    .write_at(address + 1, value.get_bits(8..=15) as u8);
+                arm.memory
+                    .write_at(address + 2, value.get_bits(16..=23) as u8);
+                arm.memory
+                    .write_at(address + 3, value.get_bits(24..=31) as u8);
+            };
+
+            self.exec_data_trasfer(reg_list, pre_post, &mut address, up_down, transfer);
```
AlessioC31, line 180: What do you think about moving this logic to `InternalMemory`? Maybe changing the trait `IoDevice`. We could introduce functions like `write_32_bits` `read_32_bits` or change the current `write_at` and `read_at` to support the amount of bytes to read/write.

guerinoni: I already thought on it, I think I will do in separated PR, I'm only thinking cleanest solution to do so :)

AlessioC31: Fair enough, can we create an issue so we don't forget? :D

guerinoni: I can do it :)

PR #105, opened 2022-11-03
ARM: Refactor `Branch`/`BranchLink` ops

AlessioC31:
I've fixed an issue with the `Branch` op since it was not sign extending the offset, so it would not have worked with negative offsets.

I also took it as an opportunity to refactor it a bit: now there's only 1 branch function instead of `branch` and `branch_and_link` that do the same thing. I've also united the tests into a single test function


guerinoni (approved):
Neat :)

PR #106, opened 2022-11-04
single data transfer ldr word

guerinoni:
- Memory: Add unused memory as map
- Everywhere: Fix lint after bumping rust version
- ARM: Implement single data transfer LDR with WORD

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -7,16 +7,16 @@ fn main() {
 
     let args = env::args().skip(1).collect::<Vec<String>>();
 
-    let cartridge_name = match args.first() {
-        Some(name) => {
-            println!("loading {name}");
-            name.clone()
-        }
-        None => {
+    let cartridge_name = args.first().map_or_else(
+        || {
             println!("no cartridge found :(");
             std::process::exit(1)
-        }
-    };
+        },
+        |name| {
+            println!("loading {name}");
+            name.clone()
+        },
+    );
 
     let options = eframe::NativeOptions {
         drag_and_drop_support: true,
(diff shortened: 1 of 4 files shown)


ui/src/cpu_inspector.rs
```diff
@@ -58,10 +58,10 @@ impl<T: Cpu> View for CpuInspector<T> {
         ui.heading("Registers");
         ui.add_space(8.0);
 
-        let registers = match self.gba.lock() {
-            Ok(gba) => gba.cpu.registers(),
-            Err(_) => vec![],
-        };
+        let registers = self
```
gallottino, line 61: Why this change? No problem about it, just for me it is more readable the first version. I don't know if there is some kind improvement.

src/main.rs
```diff
@@ -7,16 +7,16 @@ fn main() {
...
     let args = env::args().skip(1).collect::<Vec<String>>();
 
-    let cartridge_name = match args.first() {
-        Some(name) => {
-            println!("loading {name}");
-            name.clone()
-        }
-        None => {
+    let cartridge_name = args.first().map_or_else(
+        || {
```
gallottino, line 11: > Why this change? No problem about it, just for me it is more readable the first version. I don't know if there is some kind improvement

Same.

gallottino (approved):

ui/src/cpu_inspector.rs
guerinoni: More idiomatic, and also clippy complains :D

src/main.rs
guerinoni: More idiomatic, and also clippy complains :D

PR #110, opened 2022-11-06
Render in `BG_MODE_3`

gallottino:
Documentation here:

`BGCNT` [commit](https://github.com/RIP-Comm/clementine/pull/110/commits/572844925ee68d42f3b97f98df1bf5e65060d119): https://rust-console.github.io/gbatek-gbaonly/#--lcd-io-bg-control
`BG_MODE_3` [commit](https://github.com/RIP-Comm/clementine/pull/110/commits/2f840b3c502ad1aec67822f4586347a155549abe): https://rust-console.github.io/gbatek-gbaonly/#bg-mode-3---240x160-pixels-32768-colors


ui/src/utilities.rs
```diff
@@ -0,0 +1,23 @@
+use egui::Color32;
```
guerinoni, line 1: I think we can call this file `gba_color` instead of generic utilities

justfile
```diff
@@ -17,6 +17,11 @@ set positional-arguments
     @cargo build --features debug
     ./target/debug/clementine $1
 
+# run clemente in DEBUG mode
+@test_bitmap *args='':
+    @cargo build --features test_bitmap
+    ./target/debug/clementine $1
```
guerinoni, line 23: You can simply run `cargo run --features test_bitmap -- $1`

ui/Cargo.toml
```diff
@@ -8,8 +8,10 @@ license.workspace = true
 [dependencies]
 eframe = { version = "0.19.0", git = "https://github.com/emilk/egui", default-features = false, features = ["glow"] }
 egui = { version = "0.19.0", git = "https://github.com/emilk/egui", default-features = false }
-egui_extras = { version = "0.19.0", git = "https://github.com/emilk/egui" }
+egui_extras = { version = "0.19.0", git = "https://github.com/emilk/egui", features = ["image"] }
 emu = { path = "../emu"}
+image = { version = "0.24", features = ["png"] }
```
guerinoni, line 13: Maybe we can put this optional based on feature enabled

emu/src/render/lcd.rs
```diff
@@ -0,0 +1,41 @@
...
+        self[(x, y)] = color;
+    }
+}
+
+impl std::ops::Index<(usize, usize)> for Lcd {
+    type Output = Color;
+
+    fn index(&self, (x, y): (usize, usize)) -> &Color {
+        assert!(x < DISPLAY_WIDTH && y < DISPLAY_HEIGHT);
+        &self.pixels[y * DISPLAY_WIDTH + x]
```
AlessioC31, line 24: `x` are rows and `y` columns? If it is so, shouldn't it be `x * DISPLAY_WIDTH + y` ?

emu/src/render/ppu.rs
```diff
@@ -0,0 +1,175 @@
...
+        Self {
+            internal_memory,
+            lcd,
+        }
+    }
+
+    pub fn render(&self) -> Rc<RefCell<Lcd>> {
+        let gba_display = self.lcd.clone();
+
+        // BG_MODE is forced to 3 to avoid crash on start
```
AlessioC31, line 33: Will it always be like this or this is just for now that we onlt have BG_MODE 3?

emu/src/render/lcd.rs
AlessioC31: Ok, I see below that `y` is used to indicate rows and `x` to indicate columns. So this logic is correct. Is it a convention in GBA? To me, a `row, column` access is better but I guess is subjective?

emu/src/render/mod.rs
```diff
@@ -1,2 +1,27 @@
...
-pub mod palette_color;
+pub mod color;
+pub mod lcd;
+pub mod ppu;
+
+/// GBA display width
+pub const DISPLAY_WIDTH: usize = 240;
+
+/// GBA display height
+pub const DISPLAY_HEIGHT: usize = 160;
```
AlessioC31, line 10: This depends on the BG mode right? For example BG Mode 5 is `160x128` pixels

emu/src/render/lcd.rs
gallottino: Nop, x are the cols. Try to figure out an image in a cartesian plane.

emu/src/render/mod.rs
gallottino: Yes, but I want to keep it as the real size of LCD and to add for BG_MODE_5 other values referring to GBC display.

emu/src/render/lcd.rs
AlessioC31: Yup, you're right. From an image point of view it's like this. I was thinking about matrices :D

you can solve this comment

justfile
gallottino: True. I don't know why I did this ahah

emu/src/render/ppu.rs
gallottino: Just now.

emu/src/render/lcd.rs
guerinoni: We can also call `col` and `row` for better explanation :D

emu/src/gba.rs
```diff
@@ -1,21 +1,41 @@
...
-{
-    pub cpu: T,
+use crate::{
+    arm7tdmi::Arm7tdmi, cartridge_header::CartridgeHeader, memory::internal_memory::InternalMemory,
+    render::lcd::Lcd, render::ppu::PixelProcessUnit,
+};
+
+pub struct Gba {
+    pub arm_cpu: Rc<RefCell<Arm7tdmi>>,
+    //pub z80_cpu: Rc<RefCell<T>>,
```
guerinoni, line 10: Commented code

emu/src/single_data_transfer.rs
```diff
@@ -167,7 +167,7 @@ mod tests {
     #[test]
     fn check_str_byte() {
         let op_code = 0b1110_0101_1100_1111_1101_0000_0001_1000;
-        let mut cpu = Arm7tdmi::new(vec![]);
+        let mut cpu = Arm7tdmi::default();
```
guerinoni, line 170: Nice

emu/src/render/lcd.rs
gallottino: I'm not sure that for image (row, col) is a better description. I also check in other crates and (egui e.g.) and everywhere they use Image with x and y as this. If we move in other direction when we set the Image as egui do, we have a miss match of type in (x,y) tuple. 

I'm referring to GbaDisplay in ui when ColorImage is set as GbaLcd

gallottino:
Ready for merge now, I think. @AlessioC31 @guerinoni

AlessioC31 (approved):

guerinoni (approved):
Looks very good

PR #111, opened 2022-11-06
UI: Add `MemoryInspector`

AlessioC31:
Added a new UiTool to read values from memory.

I don't like very much how it is implemented but it was the only way (or at least the only one I found) to make it work since we have `Gba<T: Cpu>` and `Cpu` trait didn't have any method to access memory and memory is stored into an `Rc<RefCell<>>`.

Please tell me if you think it could be implemented in a better way (maybe without returning `Ref`? But I think it's needed since `InternalMemory` is managed by a `RefCell`).

It's very simple because I want to use it to debug an issue we may have in the instruction implementations. In the future, we could show a matrix with all the values and so on but for now I think this does the job.

### DIFF
--- a/emu/src/arm7tdmi.rs
+++ b/emu/src/arm7tdmi.rs
@@ -1,4 +1,4 @@
-use std::cell::RefCell;
+use std::cell::{Ref, RefCell};
 use std::convert::TryInto;
 use std::rc::Rc;
 
@@ -57,6 +57,7 @@ const OPCODE_ARM_SIZE: usize = 4;
 
 impl Cpu for Arm7tdmi {
     type OpCodeType = ArmModeOpcode;
+    type Memory = InternalMemory;
 
     fn fetch(&self) -> u32 {
         let instruction_index = self.registers.program_counter();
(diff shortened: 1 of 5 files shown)


gallottino:
In my refactor in #110 I removed T from GBA and I moved all the hardware outside the cpu so every UI tools can access it easly

AlessioC31:
> In my refactor in #110 I removed T from GBA and I moved all the hardware outside the cpu so every UI tools can access it easly

Yep I saw it! Now we need to decide which one should we merge first :D

I'm okay with keeping this in pause and merge yours first and then I rebase. I can use this branch in any case to debug

emu/src/arm7tdmi.rs
```diff
@@ -110,6 +111,10 @@ impl Cpu for Arm7tdmi {
     fn registers(&self) -> Vec<u32> {
         self.registers.to_vec()
     }
+
+    fn get_memory(&self) -> Ref<'_, InternalMemory> {
```
gallottino, line 115: You can return `Rc<RefCell<InternalMemory>>` instead with `self.memory.clone()` that returns another usable instance of the Rc. But I don't want to keep it for later, because cpu doesn't need to return the memory, but gba. No problem for now, just a reminder for later.

AlessioC31: Once you merge #111 and I rebase this `get_memory` fn will probably go away

AlessioC31:
I converted it to draft while we wait for #111 to be merged

guerinoni:
Agreed! We will come back to this after other PR's will be merged!

PR #112, opened 2022-11-06
Fix block data transfer

AlessioC31:
* Now correctly adds `12` when storing `R15 (PC)`
* Inverts iteration order when using decrementing. From documentation:
> The lowest register also gets transferred to/from the lowest memory address
So we should invert the order when decrementing otherwise we would store/save in inverse order.

### DIFF
--- a/emu/src/arm7tdmi.rs
+++ b/emu/src/arm7tdmi.rs
@@ -212,6 +212,7 @@ impl Arm7tdmi {
         F: Fn(&mut Self, u32, usize),
     {
         let alignment = 4; // Since are word, the alignment is 4.
+
         let change_address = |address: u32| {
             if up_down {
                 address.wrapping_add(alignment)
@@ -220,7 +221,15 @@ impl Arm7tdmi {
             }
         };
 
-        for reg_source in 0..16 {
+        // If we are decreasing we still want to store the lowest reg to the lowest
+        // memory address. For this reason we reverse the loop order.
+        let range_registers: Box<dyn Iterator<Item = u8>> = if up_down {
+            Box::new(0..=15)
+        } else {
+            Box::new((0..=15).rev())
+        };
+
+        for reg_source in range_registers {
             if reg_list.is_bit_on(reg_source) {
                 if pre_post {
                     *address = change_address(*address);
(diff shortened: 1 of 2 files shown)


emu/src/arm7tdmi.rs
```diff
@@ -316,45 +325,156 @@ mod tests {
     #[test]
     fn check_block_data_transfer() {
         {
-            // STR
-            let op_code = 0b1110_1000_1011_1101_1000_0000_0000_0111;
+            // LDM with post-increment
+            let op_code = 0b1110_100_0_1_0_1_1_1101_0000000010100010;
```
guerinoni, line 329: Don't understand why formatting the number in this way? Which part you want highlight?

emu/src/arm7tdmi.rs
```diff
@@ -221,7 +228,15 @@ impl Arm7tdmi {
...
         };
 
-        for reg_source in 0..16 {
+        // If we are decreasing we still want to store the lowest reg to the lowest
+        // memory address. For this reason we reverse the loop order.
+        let mut regs_it: Box<dyn DoubleEndedIterator<Item = u8>> = Box::new(0..16);
+
+        if !up_down {
+            regs_it = Box::new(regs_it.rev());
+        }
```
guerinoni, line 237: I think is more clear and easy something like
```rust
let range_registers = if up_down {
            0..=15
        } else {
            15..=0
        };
```

emu/src/arm7tdmi.rs
AlessioC31: To me it's a bit easier when you want to understand what this instruction does. For example in this case first 4 bits are cond, then 3 bits of opcode, 1 bit of P, 1 bit of U and so on...

AlessioC31: If you group every 4 I find it harder to see which bit does what

emu/src/arm7tdmi.rs
AlessioC31: ahahah definitely! I don't know why I went down that way

AlessioC31: Mh ok I've tried it but it looks like `15..=0` doesn't work (it's an empy range). One should do `(0..=15).rev()` but then it has a different type than `0..=15` and it doesn't work

guerinoni: Ok then we can go for
```rust
    let x: Box<dyn Iterator<Item = usize> = if b > a {
        x = Box::new(0..=15)
    } else {
        x = Box::new((0..=15).rev())
    };
```

guerinoni (approved):
Just rebase :D

AlessioC31:
Done! Should be ready @guerinoni

PR #113, opened 2022-11-06
Fix advance cond

AlessioC31:
* When the instruction can't be executed because of cond flags, the program counter should be increased instead of staying on the same instruction forever
* When printing the instruction, we should pad with leading zeroes otherwise it would not print 32 bits when `cond` starts with 0.


guerinoni (approved):
LGTM

PR #114, opened 2022-11-07
Play button

AlessioC31:
Implements `Play` button in UI.

Since UI is in a different thread everything had to be moved from `Rc<RefCell<>>` to `Arc<Mutex<>>`.

I guess there's a lot that could be improved, it's my first time working with threads in rust :D

I opted for `.unwrap()` after `.lock()` instead of `if let Ok(...) = ...lock()` because I think it's better to have panics for now instead of "silent failures"

### DIFF
--- a/ui/src/cpu_inspector.rs
+++ b/ui/src/cpu_inspector.rs
@@ -48,7 +48,14 @@ impl View for CpuInspector {
                 cartridge_name = gba.cartridge_header.game_title.clone()
             }
             ui.text_edit_singleline(&mut cartridge_name);
-            if ui.button("▶").clicked() {
+
+            if ui
+                .add_enabled(
+                    !self.play.load(std::sync::atomic::Ordering::Relaxed),
+                    egui::Button::new("▶"),
+                )
+                .clicked()
+            {
                 if self.thread_handle.is_some() {
                     return;
                 }
@@ -64,7 +71,14 @@ impl View for CpuInspector {
                     }
                 }));
             }
-            if ui.button("⏸ ").clicked() {
+
+            if ui
+                .add_enabled(
+                    self.play.load(std::sync::atomic::Ordering::Relaxed),
+                    egui::Button::new("⏸ "),
+                )
+                .clicked()
+            {
                 self.play.swap(false, std::sync::atomic::Ordering::Relaxed);
                 self.thread_handle = None;
             }
--- a/ui/src/cpu_inspector.rs
+++ b/ui/src/cpu_inspector.rs
@@ -2,16 +2,23 @@ use emu::{cpu::Cpu, gba::Gba};
 
 use crate::ui_traits::{UiTool, View};
 
+use std::sync::atomic::AtomicBool;
 use std::sync::{Arc, Mutex};
+use std::thread;
 
 pub struct CpuInspector {
     gba: Arc<Mutex<Gba>>,
-    play: bool,
+    play: Arc<AtomicBool>,
+    thread_handle: Option<thread::JoinHandle<()>>,
 }
 
 impl CpuInspector {
     pub fn new(gba: Arc<Mutex<Gba>>) -> Self {
-        Self { gba, play: false }
+        Self {
+            gba,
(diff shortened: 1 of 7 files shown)


ui/src/cpu_inspector.rs
```diff
@@ -2,16 +2,23 @@ use emu::{cpu::Cpu, gba::Gba};
...
 use crate::ui_traits::{UiTool, View};
 
+use std::sync::atomic::AtomicBool;
 use std::sync::{Arc, Mutex};
+use std::thread;
 
 pub struct CpuInspector {
     gba: Arc<Mutex<Gba>>,
-    play: bool,
+    play: Arc<AtomicBool>,
```
guerinoni, line 11: An `AtomicBool` can be shared between thread in safe way

ui/src/cpu_inspector.rs
```diff
@@ -42,11 +49,24 @@ impl View for CpuInspector {
             }
             ui.text_edit_singleline(&mut cartridge_name);
             if ui.button("▶").clicked() {
-                // Start a thread for gameboy execution
-                todo!("Start a thread for a background gameboy execution");
+                if self.thread_handle.is_some() {
```
guerinoni, line 52: It can be nice to make button "locked" for better understanding that is running :)

ui/src/cpu_inspector.rs
```diff
@@ -42,11 +49,24 @@ impl View for CpuInspector {
...
+                let gba_clone = Arc::clone(&self.gba);
+                let play_clone = Arc::clone(&self.play);
+
+                self.play.swap(true, std::sync::atomic::Ordering::Relaxed);
+
+                self.thread_handle = Some(thread::spawn(move || {
+                    while play_clone.load(std::sync::atomic::Ordering::Relaxed) {
+                        gba_clone.lock().unwrap().cpu.step();
+                    }
+                }));
```
guerinoni, line 65: Not sure about the semantic of this... The emulator work like
1. exec n steps
2. receives interrupt after MHz converted to number of step
3. refresh with the current value of "things"
4. restart

I didn't try the branch but maybe this will be out of sync with the GUI
A sketched code can be
```
loop {
   while cpu.need_irq() {
        cpu.step()
   }

   refresh_gui();
   update_other_thing();
}
```

AlessioC31: Mhh I see what you mean but for now we don’t have Irqs and we don’t even know how many cycles take each instruction

ui/src/gba_display.rs
```diff
@@ -78,12 +78,13 @@ impl View for GbaDisplay {
...
-        if let Ok(gba) = self.gba.lock() {
-            gba.ppu.render();
-            for row in 0..DISPLAY_HEIGHT {
-                for col in 0..DISPLAY_WIDTH {
-                    self.image[(col, row)] = GbaColor(gba.lcd.borrow()[(col, row)]).into();
-                }
+        let gba = self.gba.lock().unwrap();
+        let gba_lcd = gba.lcd.lock().unwrap();
+
+        gba.ppu.render();
```
gallottino, line 84:
```suggestion
        let gba = self.gba.lock().unwrap();
         gba.ppu.render();
         
        let gba_lcd = gba.lcd.lock().unwrap();
```
I suggest to change the order of these two instructions. I try to run the code but the GUI stucks in an eternal loop. I think that the reason is because `ppu.render()` tries to get a lock, but the UI alredy has it.

AlessioC31: Mh ok so for now, as I understood, UI updates every window every step, right?

With my changes since we spawn a thread that continuously lock the `gba` the other UI tools can't updated themselves, correct?

I can fix this specific case but at the moment I think we're bit far to have a correct "play loop" (cpu executes, irqs get triggered, lcd refreshed, etc...) because we miss irqs and timings and many other things.

If you're ok with it (@guerinoni, @gallottino), for now I would treat this "play" button as a fast "go to the first failure", and we come back to it once we implement more stuff in the emulator.

PR #115, opened 2022-11-09
Add new logger

guerinoni:
- Logger: Add new logger for `clementine`
- Everywhere: Use logger in the code with a special arg `--log-on-file`

Basically the feature is because I would love to analyze the data and istruction post-run of a specific ROM, since now we have enough instruction running but we don't know where are bugs :).
This allow us to log on stdout as befor of log on a file saved under `/tmp/clementine-{utc::now()}.log` with a special arg like `./clementine --log-on-file my_game.gba`.

We need to check this path on windows platform.

### DIFF
--- a/logger/src/lib.rs
+++ b/logger/src/lib.rs
@@ -1,0 +1,135 @@
+use std::{
+    fs::File,
+    io::{self, Write},
+    sync::Mutex,
+    time::Instant,
+};
+
+use chrono::Utc;
+use once_cell::sync::OnceCell;
+
+static LOGGER: OnceCell<Logger> = OnceCell::new();
+
+struct LoggerImpl {
+    pub sink: Box<dyn Write + Send>,
+    pub start_instant: Instant,
+}
+
+impl LoggerImpl {
+    fn new(kind: LogKind) -> Self {
+        let start_instant = Instant::now();
+        match kind {
+            LogKind::STDOUT => Self {
+                sink: Box::new(io::stdout()),
+                start_instant,
+            },
+            LogKind::FILE => {
+                let now = Utc::now();
+                let filename = format!("clementine-{}.log", now.timestamp());
+                let path = std::env::temp_dir().join(filename);
+                Self {
+                    sink: Box::new(File::create(path).unwrap()),
+                    start_instant,
+                }
+            }
+        }
+    }
+
+    fn log<T>(&mut self, data: T)
+    where
+        T: std::fmt::Display,
+    {
+        let now = self.start_instant.elapsed();
+        let seconds = now.as_secs();
+        let hours = seconds / 3600;
+        let minutes = (seconds / 60) % 60;
+        let seconds = seconds % 60;
+        let miliseconds = now.subsec_millis();
+
+        let _ = writeln!(
+            self.sink,
+            "[{:02}:{:02}:{:02}.{:03}] {}",
+            hours, minutes, seconds, miliseconds, data
+        );
+    }
+}
+
+/// LogKind represents the kind of logging: `stdout` or `logfile`.
+#[derive(PartialEq, Eq)]
+pub enum LogKind {
+    /// It logs to console, the default choice.
+    STDOUT,
+
+    /// It logs on a file in /tmp/clementine-<timestamp>.log
+    FILE,
+}
+
+/// Logger
+struct Logger {
+    pub inner_impl: Mutex<LoggerImpl>,
+}
+
+impl Default for Logger {
+    fn default() -> Self {
+        Self {
+            inner_impl: Mutex::new(LoggerImpl::new(LogKind::STDOUT)),
+        }
+    }
+}
+
+impl Logger {
+    fn new(kind: LogKind) -> Self {
+        Self {
+            inner_impl: Mutex::new(LoggerImpl::new(kind)),
+        }
+    }
+
+    fn log<T>(&self, data: T)
+    where
+        T: std::fmt::Display,
+    {
+        if let Ok(ref mut inner) = self.inner_impl.lock() {
+            inner.log(data);
+        }
+    }
+}
+
+pub fn init_logger(kind: LogKind) {
+    _ = LOGGER.set(Logger::new(kind));
+}
+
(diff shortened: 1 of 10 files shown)


logger/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+    fn new(kind: LogKind) -> Self {
+        let start_instant = Instant::now();
+        match kind {
+            LogKind::STDOUT => Self {
+                sink: Box::new(io::stdout()),
+                start_instant,
+            },
+            LogKind::FILE => {
+                let now = Utc::now();
+                let path = format!("/tmp/clementine-{}-{}.log", now.date(), now.time());
```
AlessioC31, line 28: I would use something like https://doc.rust-lang.org/std/env/fn.temp_dir.html

AlessioC31 (changes requested):
Curiosity: aren’t there in rust logging solutions instead of writing our own?

AlessioC31:
> Curiosity: aren’t there in rust logging solutions instead of writing our own?
> 
> somethin like https://docs.rs/log/latest/log/
> 
> so that we could have different levels of logging, for example I imagine a level where we would also get the register dump every instruction and so on

I would also squash first and third commit

guerinoni:
> Curiosity: aren’t there in rust logging solutions instead of writing our own?
> 
> somethin like https://docs.rs/log/latest/log/
> 
> so that we could have different levels of logging, for example I imagine a level where we would also get the register dump every instruction and so on

I will squash commits... Basically it is enough easy to implement myself, also for personal fun (“What I cannot create, I do not understand.” – [Richard Feynman](https://en.m.wikiquote.org/wiki/Richard_Feynman))

Another reason is because this will never be a production application, one day, when it will run it will be finished because there are no other rooms to improve or create new game around it, then we are just creating fun for ourselves :D

AlessioC31:
> > Curiosity: aren’t there in rust logging solutions instead of writing our own?
> > somethin like https://docs.rs/log/latest/log/
> > so that we could have different levels of logging, for example I imagine a level where we would also get the register dump every instruction and so on
> 
> I will squash commits... Basically it is enough easy to implement myself, also for personal fun (“What I cannot create, I do not understand.” – [Richard Feynman](https://en.m.wikiquote.org/wiki/Richard_Feynman))
> 
> Another reason is because this will never be a production application, one day, when it will run it will be finished because there are no other rooms to improve or create new game around it, then we are just creating fun for ourselves :D

Although I like and enforce Feynman’s quote in my life too (that’s why we’re doing this emulator when there exist thousands already :D), I don’t agree with reimplementing something as standardised as a logging library in a repo where the main focus is something else, because then you need to draw boundaries somewhere: e.g., can person X implement from scratch in clementine a new UI library? Why not? What we define simple enough to be reimplemented here and what we don’t?

AlessioC31:
> > > > Curiosity: aren’t there in rust logging solutions instead of writing our own?
> > > > somethin like https://docs.rs/log/latest/log/
> > > > so that we could have different levels of logging, for example I imagine a level where we would also get the register dump every instruction and so on
> > > 
> > > 
> > > I will squash commits... Basically it is enough easy to implement myself, also for personal fun (“What I cannot create, I do not understand.” – [Richard Feynman](https://en.m.wikiquote.org/wiki/Richard_Feynman))
> > > Another reason is because this will never be a production application, one day, when it will run it will be finished because there are no other rooms to improve or create new game around it, then we are just creating fun for ourselves :D
> > 
> > 
> > Although I like and enforce Feynman’s quote in my life too (that’s why we’re doing this emulator when there exist thousands already :D), I don’t agree with reimplementing something as standardised as a logging library in a repo where the main focus is something else, because then you need to draw boundaries somewhere: e.g., can person X implement from scratch in clementine a new UI library? Why not? What do we define simple enough to be reimplemented here and what we don’t?
> 
> No blockers from my side, we can implement all from scratch, sure there is enough difference from a single file for logger or entire UI package, but can be interesting re-write a UI specific for our use case :)

I do like the concept that everyone can code everything from scratch (who am I to say no?) but I just think it should not be under clementine since the repo is for the emulator

logger/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+
+pub fn log<T>(data: T)
+where
+    T: std::fmt::Display,
+{
+    LOGGER.get().map_or((), |logger| logger.log(data));
+}
+
+#[cfg(test)]
+mod tests {
```
FedericoBruzzone, line 108: Looks good.
Quick question, to you think to modify the tests to make them pass on windows? (It’s no longer my case, but maybe someone runs it on windows)

FedericoBruzzone (approved):

logger/src/lib.rs
guerinoni: Done

logger/src/lib.rs
guerinoni: Sure

FedericoBruzzone (approved):

logger/src/lib.rs
FedericoBruzzone: Just for information, what was the problem?

AlessioC31 (approved):

logger/src/lib.rs
```diff
@@ -0,0 +1,135 @@
...
+
+        let _ = writeln!(
+            self.sink,
+            "[{:02}:{:02}:{:02}.{:03}] {}",
+            hours, minutes, seconds, miliseconds, data
+        );
+    }
+}
+
+/// LogKind represents the kind of logging are available now.
```
AlessioC31, line 57:
```suggestion
/// LogKind represents the kind of logging: `stdout` or `logfile`.
```

AlessioC31 (changes requested):

logger/src/lib.rs
guerinoni: Creating a file in tmp directory using `:` cause a strange error on windows, now the file is called `clementine-<timestamp>.log` and before I have used a date with time like `12:00`
Another observation is that in unix system we use `/tmp` for temporary data instead under windows is `C:/users/<name>/appdata/temp`

PR #117, opened 2022-11-12
ARM: Improve decode instruction from op_code (u32)

guerinoni:
Aligned all instruction with the document of proper CPU. Many of those
are not implemented, neither displayed well in formatted log, but now we
have all istruction decode well (I hope :')).
Now is more easy to understand which bits should be true/false, using
mask were not so easy to understand from the log.
The decode function is ordered on how many bit are already know as
invariant.

Screenshot 2022-11-12 at 16 33 35

### DIFF
--- a/emu/src/arm7tdmi.rs
+++ b/emu/src/arm7tdmi.rs
@@ -88,7 +88,7 @@ impl Cpu for Arm7tdmi {
             SingleDataTransfer => self.single_data_transfer(op_code),
             Undefined => todo!(),
             BlockDataTransfer => self.block_data_transfer(op_code),
-            Branch => self.branch(op_code, true), // FIXME: Handle link inside this function.
+            Branch => self.branch(op_code),
             CoprocessorDataTransfer => todo!(),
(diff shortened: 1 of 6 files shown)


emu/src/arm7tdmi.rs
```diff
@@ -82,14 +82,12 @@ impl Cpu for Arm7tdmi {
         // Instruction functions should return whether PC has to be advanced
         // after instruction executed.
         let should_advance_pc = match op_code.instruction {
-            Branch => self.branch(op_code, false),
-            BranchLink => self.branch(op_code, true),
             DataProcessing1 | DataProcessing2 | DataProcessing3 => self.data_processing(op_code),
+            BranchAndBranchWithLink => self.branch(op_code, true),
+            BranchAndExchange => self.branch(op_code, false),
```
AlessioC31, line 87: I'd put `unimplemented!` here since it's `BX` is different than a regular `branch`

AlessioC31 (changes requested):

FedericoBruzzone (approved):

Issue #100, opened 2022-10-29
PR CI: Add check on rebase


AlessioC31:
We currently have checks (fmt, test, lint) on PR branches. It would be useful to execute the same set of checks on the "potential" new `main` branch (PR branch rebased to current main).

This is useful because the PR branch may be not rebased to the `main`, thus merging the PR could result in broken tests.

AlessioC31:
Another option could be to disable merging the PR if it is not rebased on `main`. If we go for this we could enable some actions that automatically rebase PRs when you comment "/rebase" or something like that.

Let's use this issue to discuss :)

@guerinoni

guerinoni:
Something simple like

```bash
command:
      - |
          before=\$(git rev-parse HEAD)
          echo "Before: \$before"
          echo Rebasing main into PR branch...
          git rebase main --message 'Automatic rebase main for PR branch'
          after=\$(git rev-parse HEAD)
          echo "After: \$after"

          [[ \$before == \$after ]] || run_test
```

AlessioC31:
Closing since it looks like merge queues solve this problem

PR #118, opened 2022-11-12
CI: Execute pipeline in rebased PR

AlessioC31:
Closes: #100

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -20,17 +20,7 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
-  fmt:
-    runs-on: ubuntu-latest
-
-    steps:
-    - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -20,17 +20,24 @@ env:
...
+    - uses: dtolnay/rust-toolchain@stable
+    - name: Check
+      run: |
+        just check-fmt
+        just test
+        just lint
 
-  test:
-    needs: [fmt]
+  check_after_rebase:
```
guerinoni, line 40: We can keep `needs: [check]` at least for avoid useless check

guerinoni:
Ok LGTM, but I have a suggestion on this... We can keep the file as it is in `main` and just adding `check-after-rebase` at the end, in this way we can check pipeline is working, after this check we can refactor the file in another moment

FedericoBruzzone (approved):

PR #119, opened 2022-11-13
Add render in BG_MODE_4

gallottino:
Documented here: [BG_MODE_4](https://rust-console.github.io/gbatek-gbaonly/#bg-mode-4---240x160-pixels-256-colors-out-of-32768-colors)

In this PR I also added a macro to simplify lock syntax:

`acquire_lock!(lock_var, acquired => { you can use acquired here } );`

It was usefull because sometime we have to acquire a lock only in the needed scopes. In this way it is more cleaner to use a mutex more times in the same flow. 

For example assume you have two function `fn_1()`, `fn_2()`

```
fn_1() {
  // it needs a lock
  let acquired_var = self.lock_var.lock().unwrap();
  acquired_var.do_something();

 fn_2();
}
```

```
fn_2() {
  // it needs a lock
  let acquired_var = self.lock_var.lock().unwrap();
  acquired_var.do_other_stuff();
}
```
if you call `fn_1()` a deadlock occurs.

with this macro you can do this without problems:
```
fn_1() {
  // it needs a lock
  acquire_lock!(self.lock_var, var => { var.do_something(); });
 fn_2();
}
```

With `just test_mode_4 path/rom.gba` you can check the correct behavior with the help of `PaletteVisualizer`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ target/
 
 # Support for Project snippet scope
 .vscode/*.code-snippets
+.vscode/
 
 # Ignore code-workspaces
 *.code-workspace
--- a/emu/src/render/ppu.rs
+++ b/emu/src/render/ppu.rs
@@ -1,7 +1,11 @@
+use macros::acquire_lock;
 use std::sync::{Arc, Mutex};
 
 #[cfg(feature = "debug")]
 use rand::Rng;
+
+#[cfg(feature = "debug")]
+use crate::render::color::colors;
 
 use crate::{
     memory::internal_memory::InternalMemory,
@@ -31,11 +35,9 @@ impl PixelProcessUnit {
     }
 
     pub fn render(&self) {
-        let memory = self.internal_memory.lock().unwrap();
-        let mut gba_lcd = self.gba_lcd.lock().unwrap();
-
         #[allow(unused_assignments)]
-        let mut bg_mode = memory.lcd_registers.get_bg_mode();
+        let mut bg_mode =
+            acquire_lock!(self.internal_memory, memory => { memory.lcd_registers.get_bg_mode() });
 
         // BG_MODE_3 forced for now.
         bg_mode = 3;
@@ -43,6 +45,11 @@ impl PixelProcessUnit {
         #[cfg(feature = "mode_3")]
         {
             bg_mode = 3;
+        }
+
+        #[cfg(feature = "mode_4")]
+        {
+            bg_mode = 4;
         }
 
         #[cfg(feature = "mode_5")]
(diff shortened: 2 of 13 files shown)


ui/src/cpu_inspector.rs
```diff
@@ -67,7 +71,8 @@ impl View for CpuInspector {
 
                 self.thread_handle = Some(thread::spawn(move || {
                     while play_clone.load(std::sync::atomic::Ordering::Relaxed) {
-                        gba_clone.lock().unwrap().cpu.step();
+                        thread::sleep(Duration::from_millis(1000 / FPS));
+                        acquire_lock!(gba_clone, gba => {gba.step()});
```
AlessioC31, line 75: I'm not sure about the semantics. Does 1 fps corresponds to a cpu cycle + lcd refresh?

emu/src/render/ppu.rs
```diff
@@ -31,11 +35,9 @@ impl PixelProcessUnit {
     }
 
     pub fn render(&self) {
-        let memory = self.internal_memory.lock().unwrap();
-        let mut gba_lcd = self.gba_lcd.lock().unwrap();
-
         #[allow(unused_assignments)]
-        let mut bg_mode = memory.lcd_registers.get_bg_mode();
+        let mut bg_mode =
+            acquire_lock!(self.internal_memory, memory => { memory.lcd_registers.get_bg_mode() });
```
AlessioC31, line 40: I think we could only lock once but I guess it's not a big deal

emu/src/render/ppu.rs
```diff
@@ -61,42 +68,74 @@ impl PixelProcessUnit {
...
+                            (y * LCD_WIDTH + x) + (selected_frame * LCD_HEIGHT * LCD_WIDTH);
+
+                        let index_palette = acquire_lock!(self.internal_memory, memory => {
+                            memory.video_ram[index_mem]
+                        });
+
+                        let color: Color = self
+                            .get_color_from_full_palette(index_palette.into(), &PaletteType::BG);
+
+                        acquire_lock!(self.gba_lcd, gba_lcd => { gba_lcd.set_pixel(x, y, color); })
```
AlessioC31, line 108: Can we lock outside of the for so that we don't lock every iteration? I think it's also semantically more correct because we lock the access to memory/lcd while we refresh it and then we release it

emu/src/render/ppu.rs
```diff
@@ -61,42 +68,74 @@ impl PixelProcessUnit {
...
+
+                            let color_array =
+                                [
+                                    memory.video_ram[index_color],
+                                    memory.video_ram[index_color + 1],
+                                ];
+
+                            let color: Color = [color_array[0], color_array[1]].into();
+
+                            acquire_lock!(self.gba_lcd, gba_lcd => { gba_lcd.set_gbc_pixel(x, y, color); })
```
AlessioC31, line 135: Same as above

emu/src/render/ppu.rs
```diff
@@ -61,42 +68,74 @@ impl PixelProcessUnit {
...
+                            let index_color = (y * LCD_WIDTH + x) * 2;
+
+                            let color_array = [
+                                memory.video_ram[index_color],
+                                memory.video_ram[index_color + 1],
+                            ];
+
+                            let color: Color = [color_array[0], color_array[1]].into();
+
+                            acquire_lock!(self.gba_lcd, gba_lcd => { gba_lcd.set_pixel(x, y, color); })
```
AlessioC31, line 84: Same as below (lock outside for)

emu/src/render/ppu.rs
```diff
@@ -61,42 +68,74 @@ impl PixelProcessUnit {
...
+                        }
+                    }
+                });
+            }
+            4 => {
+                // 06000000-06009FFF for Frame 0
+                // 0600A000-06013FFF for Frame 1
```
AlessioC31, line 91: Why `06009FFF` and `06013FFF` instead of `060095FF` and `060135FF` as documentation says?

emu/src/render/ppu.rs
```diff
@@ -61,42 +68,74 @@ impl PixelProcessUnit {
...
-                            memory.video_ram[index_color],
-                            memory.video_ram[index_color + 1],
-                        ]
-                        .into();
-
-                        gba_lcd.set_pixel(x, y, color);
+                        let index_mem =
+                            (y * LCD_WIDTH + x) + (selected_frame * LCD_HEIGHT * LCD_WIDTH);
+
+                        let index_palette = acquire_lock!(self.internal_memory, memory => {
```
AlessioC31, line 101: Same as below (lock outside for)

AlessioC31 (changes requested):
The commit is a bit big, what do you think about splitting it? (1 for adding `acquire_lock`, 1 for `bg_mode 4`, 1 for fps+gba step?)

emu/src/render/ppu.rs
```diff
@@ -61,42 +68,74 @@ impl PixelProcessUnit {
...
-                        let index_color = (y * LCD_WIDTH + x) * 2;
-                        let color: Color = [
-                            memory.video_ram[index_color],
-                            memory.video_ram[index_color + 1],
-                        ]
-                        .into();
-
-                        gba_lcd.set_pixel(x, y, color);
+                        let index_mem =
+                            (y * LCD_WIDTH + x) + (selected_frame * LCD_HEIGHT * LCD_WIDTH);
```
AlessioC31, line 99: I'm not sure this is correct, one frame is `0x9600` bytes long so the first one goes from `0x06000000` to `0x060095FF`, the second one starts at `0600A000`, with this `index_mem` if I were to select pixel `0;0` of Frame 1 I would have:

`0 * LCD_WIDTH + 0` + (1 * 0x9600) = 0x9600` so when accessing `video_ram` I would read from `0x6009600` which is incorrect since the Frame 1 starts from `0x600A000`

AlessioC31 (changes requested):

ui/src/cpu_inspector.rs
gallottino: It is not correct. It put it just to avoid  GUI slowdown. We need to manage this thread, because I think that it is better put in pause the execution when an user is in the GUI.  Maybe I'll address it later. For now it is an escamotage.

emu/src/render/ppu.rs
gallottino: Yeah. I had the same idea, but `get_color_from_full_palette()` in `BG_MODE_4` uses the same lock.

emu/src/render/ppu.rs
gallottino: The problem is the method in line 106.

emu/src/render/ppu.rs
gallottino: Yep, you are right. I assumed that the frame are contiguous.

emu/src/render/ppu.rs
gallottino: Wrong copy paste. They are the GBC frames.

emu/src/render/ppu.rs
gallottino: I also noticed that for MODE_5 I did the same error ahaha

gallottino: I also noticed that for BG_MODE_5 I did the same error :( ahah

emu/src/render/ppu.rs
AlessioC31: Mhh I get it but I don't know, I would try to rethink the why it works. I know we should not focus on performance right now but I think it's better to address this now that we're very early stage

gallottino: I'll think to a solution so.

.gitignore
```diff
@@ -31,6 +31,7 @@ target/
 
 # Support for Project snippet scope
 .vscode/*.code-snippets
+.vscode/
```
guerinoni, line 34: Also this can be a different atomic commit

macros/src/utility_macros.rs
```diff
@@ -0,0 +1,10 @@
+#[macro_export]
+macro_rules! acquire_lock {
+    ($mutex:expr, $lock:ident => $exec:block ) => {
+        match $mutex.lock() {
+            #[allow(unused_mut)]
+            Ok(mut $lock) => $exec,
+            _ => Default::default(),
+        }
+    };
+}
```
guerinoni, line 10: I would love to see some tests here :)

emu/src/render/ppu.rs
guerinoni: Good catch

emu/src/render/ppu.rs
guerinoni: This macro only add a strict scope of the locks? Because maybe we can achieve this in other way if it is the only limit of our design now

PR #120, opened 2022-11-16
ARM: Implement Halfword data transfer immediate offset

guerinoni:

### DIFF
--- a/emu/src/arm7tdmi.rs
+++ b/emu/src/arm7tdmi.rs
@@ -84,7 +84,7 @@ impl Cpu for Arm7tdmi {
             SingleDataSwap => todo!(),
             BranchAndExchange => todo!(),
             HalfwordDataTransferRegisterOffset => todo!(),
-            HalfwordDataTransferImmediateOffset => todo!(),
+            HalfwordDataTransferImmediateOffset => self.data_transfer_immediate_offset(op_code),
             SingleDataTransfer => self.single_data_transfer(op_code),
             Undefined => todo!(),
             BlockDataTransfer => self.block_data_transfer(op_code),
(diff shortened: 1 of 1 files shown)


emu/src/arm7tdmi.rs
```diff
@@ -122,6 +122,64 @@ impl Arm7tdmi {
...
+        let immediate_offset_low = op_code.get_bits(0..=3);
+
+        let offset = (immediate_offset_high << 4) | immediate_offset_low;
+        let mut address = rn_base_register;
+
+        if rn_base_register == 0xF {
+            // prefetching
+            let v: u32 = self.registers.program_counter().try_into().unwrap();
+            address = address.wrapping_add(v + 8);
+        }
```
AlessioC31, line 143: When `rn != 0xF` we're not reading its value, I would do something like:

```rust
let mut address = self.registers.register_at(rn_base_register.try_into().unwrap());

if rn_base_register == 0xF {
    address = address.wrapping_add(address + 8);
}
```

emu/src/arm7tdmi.rs
```diff
@@ -122,6 +122,64 @@ impl Arm7tdmi {
...
+                _ => unreachable!("HS flags can't be != from 01 for LOAD (L)=0"),
+            };
+        }
+
+        if !pre_post {
+            // TODO: ignore write back (should be 0 in this case but...)
+            todo!()
+        }
+
+        true
```
AlessioC31, line 180: if we load into `R15` we should return `false`, I know this is not the case but since it's a trivial check I would add it now so that we don't forget

emu/src/arm7tdmi.rs
```diff
@@ -122,6 +122,64 @@ impl Arm7tdmi {
...
+            };
+
+            match transfer_type {
+                HalfwordTransferType::UnsignedHalfwords => {
+                    if let Ok(mut mem) = self.memory.lock() {
+                        mem.write_at(address, value.get_bits(0..=7) as u8);
+                        mem.write_at(address + 1, value.get_bits(8..=15) as u8);
+                    }
+                }
+                _ => unreachable!("HS flags can't be != from 01 for LOAD (L)=0"),
```
AlessioC31, line 171:
```suggestion
                _ => unreachable!("HS flags can't be != from 01 for STORE (L=0)"),
```

AlessioC31 (changes requested):

emu/src/arm7tdmi.rs
guerinoni: Right!

emu/src/arm7tdmi.rs
guerinoni: Fixed!

emu/src/arm7tdmi.rs
guerinoni: Eheheh easy to confuse with all those flags

AlessioC31 (changes requested):

emu/src/arm7tdmi.rs
```diff
@@ -122,6 +122,66 @@ impl Arm7tdmi {
...
+                _ => unreachable!("HS flags can't be != from 01 for STORE (L=0)"),
+            };
+        }
+
+        if !pre_post {
+            // TODO: ignore write back (should be 0 in this case but...)
+            todo!()
+        }
+
+        rd_source_destination_register != 0xF
```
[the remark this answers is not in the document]
guerinoni: Don't remember where we red this but I trust you :D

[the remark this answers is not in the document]
AlessioC31: > Don't remember where we red this but I trust you :D

Wait it's probably wrong ahahah

We should return false only if we're loading to R15 so:

`!(load_store && rd_source_destination_register == 0xF)`

Sorry!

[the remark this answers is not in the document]
guerinoni: Eheheh my doubt are worth 😂

AlessioC31 (approved):

PR #121, opened 2022-11-17
`spsr` and register bank

AlessioC31:
* Refactored `Cpsr` to `Psr` to use the same struct for both `cpsr` and `spsr`
* Added `RegisterBank` struct into `Armv7tdmi`. This struct stores the banked registers. Banked general purpose registers (R8-R14 for `fiq` and R13-R14 for `supervisor`, `abort`, `irq`, and `undefined`) are mapped to general purpose registers when in the given mode. For example: when in mode `fiq` if the CPU accesses R13, it's using the `r13_fiq`. For this reason, I've added `bank_registers` to save banked registers depending on the current mode and `restore_registers` to restore from the bank.

This work is preliminary to implementing `PSR` instructions and to fix all the instructions already implemented to take into account spsr registers in some specific cases

I'm not 100% sure the idea I came up with the `RegisterBank` is the best one, we can discuss it here and in any case, we'll see once we'll start using this machinery :D

Documentation: http://bear.ces.cwru.edu/eecs_382/ARM7-TDMI-manual-pt1.pdf pages from 3-4 to to 3-10 (except THUMB sections)

### DIFF
--- a/emu/src/arm/arm7tdmi.rs
+++ b/emu/src/arm/arm7tdmi.rs
@@ -7,6 +7,7 @@ use crate::arm::cpu_modes::Mode;
 use crate::arm::instruction::ArmModeInstruction;
 use crate::arm::opcode::ArmModeOpcode;
 use crate::arm::psr::Psr;
+use crate::arm::register_bank::RegisterBank;
 use crate::bitwise::Bits;
 use crate::cpu::Cpu;
 use crate::memory::internal_memory::InternalMemory;
@@ -50,6 +51,8 @@ pub struct Arm7tdmi {
 
     pub cpsr: Psr,
     pub registers: Registers,
+
+    pub register_bank: RegisterBank,
 }
 
 impl Default for Arm7tdmi {
@@ -59,6 +62,7 @@ impl Default for Arm7tdmi {
             memory: Arc::new(Mutex::new(InternalMemory::default())),
             cpsr: Psr::from(Mode::User), // FIXME: Starting as User? Not sure
             registers: Registers::default(),
+            register_bank: RegisterBank::default(),
         }
     }
 }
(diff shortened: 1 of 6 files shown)


emu/src/arm/arm7tdmi.rs
```diff
@@ -43,15 +44,28 @@ impl Registers {
...
 
+impl Default for Arm7tdmi {
+    fn default() -> Self {
+        // SPSR are initalized with the value related to their respective
+        // mode (5 least significant bits).
+        // e.g., FIQ SPSR has mode 0b10001, IRQ 0b10010, etc...
+        Self {
+            rom: Arc::new(Mutex::new(Vec::default())),
+            memory: Arc::new(Mutex::new(InternalMemory::default())),
+            cpsr: Psr::from(Mode::User), // FIXME: Starting as User? Not sure
```
guerinoni, line 63: cpsr -> psr

emu/src/arm/arm7tdmi.rs
```diff
@@ -50,18 +51,18 @@ pub struct Arm7tdmi {
...
     fn default() -> Self {
-        // SPSR are initalized with the value related to their respective
-        // mode (5 least significant bits).
-        // e.g., FIQ SPSR has mode 0b10001, IRQ 0b10010, etc...
         Self {
             rom: Arc::new(Mutex::new(Vec::default())),
             memory: Arc::new(Mutex::new(InternalMemory::default())),
             cpsr: Psr::from(Mode::User), // FIXME: Starting as User? Not sure
             registers: Registers::default(),
+            register_bank: RegisterBank::default(),
```
guerinoni, line 65: Maybe we can `..Default::default()`

emu/src/arm/psr.rs
```diff
@@ -2,7 +2,7 @@ use crate::arm::{condition::Condition, cpu_modes::Mode, data_processing::Arithme
 use crate::bitwise::Bits;
 
 /// Program Status Register.
-#[derive(Default)]
+#[derive(Default, Clone)]
```
guerinoni, line 5: We can also add `Copy` since is a built in type, no wasted copy

emu/src/arm/arm7tdmi.rs
```diff
@@ -43,13 +45,26 @@ impl Registers {
...
+}
+
+impl Default for Arm7tdmi {
+    fn default() -> Self {
+        Self {
+            rom: Arc::new(Mutex::new(Vec::default())),
+            memory: Arc::new(Mutex::new(InternalMemory::default())),
+            cpsr: Psr::from(Mode::User), // FIXME: Starting as User? Not sure
+            registers: Registers::default(),
+            register_bank: RegisterBank::default(),
```
guerinoni, line 65: Can we use `..Default::default()`?

emu/src/arm/arm7tdmi.rs
```diff
@@ -183,6 +197,97 @@ impl Arm7tdmi {
         !(load_store && rd_source_destination_register == 0xF)
     }
 
+    /// Stores the banked registers of the current mode to the register bank.
+    /// It should be used before leaving the current mode.
+    pub fn bank_registers(&mut self) {
```
guerinoni, line 202: Is it a better name `store_bank_registers`?

emu/src/arm/cpu_modes.rs
```diff
@@ -0,0 +1,33 @@
...
+    System = 0b11111,
+}
+
+impl From<Mode> for u32 {
+    fn from(m: Mode) -> Self {
+        m as Self
+    }
+}
+
+impl TryFrom<u32> for Mode {
```
guerinoni, line 18: Why both implementation? (from and tryfrom)

emu/src/arm/arm7tdmi.rs
AlessioC31: I tried but then it's recursive :(

emu/src/arm/arm7tdmi.rs
AlessioC31: I think `cpsr` is okay since this variable in particular is the current psr

emu/src/arm/psr.rs
AlessioC31: Sure, thanks!

emu/src/arm/arm7tdmi.rs
AlessioC31: What do you think about `store_registers_in_bank`?

emu/src/arm/cpu_modes.rs
AlessioC31: One is from `Mode` to `u32` and one try from `u32` to `Mode`.

Do you think they are redundant?

emu/src/arm/arm7tdmi.rs
AlessioC31: I tried but then it's recursive :(

emu/src/arm/cpu_modes.rs
guerinoni: For me we can just add what we need, also from and tryfrom are different because for first case you will do `into()` (no failure)... Instead a tryfrom requires a `try_into()` with a possible failure

emu/src/arm/arm7tdmi.rs
guerinoni: Is good, just use a verb with an action for what we mean :D

emu/src/arm/arm7tdmi.rs
guerinoni: Oh yes yes, we are implementing default here :)

emu/src/arm/cpu_modes.rs
guerinoni: Ok make sense

emu/src/arm/arm7tdmi.rs
AlessioC31: Done thanks!

guerinoni (approved):
LGTM

AlessandroGrassi99 (approved):
Everything seems to be okay for me

PR #123, opened 2022-11-18
Add pre-commit hook

AlessioC31:
Added `pre-commit` git hook and instructions on how to activate it.
I decided to not use `just` in the pre-commit hook since it may be that developers not use it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,3 +40,13 @@ Nobody is perfect, and sometimes we mess things up. That said, here are some goo
 * End commit message subject lines with a period.
 * Include commented-out code.
 * Engage in excessive "feng shui programming" by moving code around without quantifiable benefit.
+
+### Pre-commit hooks
+
+We use the the `pre-commit` git hook to enforce that every commit passes fmt/lint/test checks.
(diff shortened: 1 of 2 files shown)


extra/precommit-script.sh
```diff
@@ -0,0 +1,15 @@
+#/bin/sh
```
guerinoni, line 1: `#! /bin/sh`

extra/precommit-script.sh
```diff
@@ -0,0 +1,15 @@
...
+
+cargo clippy --workspace -- -D warnings         \
+    -W clippy::complexity                       \
+    -W clippy::correctness                      \
+    -W clippy::nursery                          \
+    -W clippy::perf                             \
+    -W clippy::style                            \
+    -W clippy::suspicious
+
+cargo test --workspace
```
guerinoni, line 15: Missing end empty line :)

extra/precommit-script.sh
AlessioC31: > `#! /bin/sh`

Whoops, thanks

guerinoni (approved):
LGTM

PR #129, opened 2022-12-01
ROM in Memory

guerinoni:
- Memory: Use for address `usize` type

This because in most of place is more convenient keep conversion outside the memory functions.

- CPU+Memory: Move rom data into memory

Now rom data is mapped into memory struct, that's because PC starts from 0 and read/write from bios.
This means that now we need bios for start a game.

### DIFF
--- a/emu/src/memory/internal_memory.rs
+++ b/emu/src/memory/internal_memory.rs
@@ -1,5 +1,4 @@
 use std::collections::HashMap;
-use std::sync::{Arc, Mutex};
 
 use logger::log;
 
@@ -35,7 +34,13 @@ pub struct InternalMemory {
 
     /// From 0x08000000 to 0x0FFFFFFF.
     /// Basically here you can find different kind of rom loaded.
-    pub rom: Arc<Mutex<Vec<u8>>>,
+    // TODO: Not sure if we should split this into
+    // 08000000-09FFFFFF Game Pak ROM/FlashROM (max 32MB) - Wait State 0
+    // 0A000000-0BFFFFFF Game Pak ROM/FlashROM (max 32MB) - Wait State 1
+    // 0C000000-0DFFFFFF Game Pak ROM/FlashROM (max 32MB) - Wait State 2
+    // 0E000000-0E00FFFF Game Pak SRAM (max 64 KBytes) - 8bit Bus width
+    // 0E010000-0FFFFFFF Not used
+    pub rom: Vec<u8>,
 
     /// From 0x00004000 to 0x01FFFFFF.
     /// From 0x10000000 to 0xFFFFFFFF.
@@ -44,12 +49,12 @@ pub struct InternalMemory {
 
 impl Default for InternalMemory {
     fn default() -> Self {
-        Self::new([0_u8; 0x00004000], Arc::new(Mutex::new(vec![])))
+        Self::new([0_u8; 0x00004000], vec![])
     }
 }
 
 impl InternalMemory {
-    pub fn new(bios: [u8; 0x00004000], rom: Arc<Mutex<Vec<u8>>>) -> Self {
+    pub fn new(bios: [u8; 0x00004000], rom: Vec<u8>) -> Self {
         Self {
             bios_system_rom: bios.into(),
             working_ram: vec![0; 0x00040000],
--- a/emu/src/memory/internal_memory.rs
+++ b/emu/src/memory/internal_memory.rs
@@ -58,7 +58,7 @@ impl InternalMemory {
         }
     }
 
-    fn write_address_lcd_register(&mut self, address: u32, value: u8) {
+    fn write_address_lcd_register(&mut self, address: usize, value: u8) {
         match address {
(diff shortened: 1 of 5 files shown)


emu/src/memory/internal_memory.rs
```diff
@@ -209,7 +215,7 @@ impl InternalMemory {
 }
 
 impl IoDevice for InternalMemory {
-    type Address = u32;
+    type Address = usize;
     type Value = u8;
 
     fn read_at(&self, address: Self::Address) -> Self::Value {
```
AlessioC31, line 221: We should add the case when we read from the ROM mapped area, right? (The one starting at `0x08...`)

AlessioC31, line 221: We should add the case when we read from the ROM mapped area, right? (The one starting at `0x08...`)

AlessioC31 (changes requested):

emu/src/memory/internal_memory.rs
```diff
@@ -29,8 +33,9 @@ pub struct InternalMemory {
     /// From 0x06000000 to 0x06017FFF (96 kb).
     pub video_ram: Vec<u8>,
 
-    /// From 0x04000100 to 0x0400010E.
-    timer_registers: TimerRegisters,
+    /// From 0x08000000 to 0x0FFFFFFF.
+    /// Basically here you can find different kind of rom loaded.
+    pub rom: Arc<Mutex<Vec<u8>>>,
```
AlessioC31, line 38: Do we need the rom to be wrapped around `Arc<Mutex<>>` ?

AlessioC31 (changes requested):

emu/src/memory/internal_memory.rs
guerinoni: I can add it :)

emu/src/memory/internal_memory.rs
guerinoni: Don't think we need now, I think it was like this because of some "viewer" to create in UI then should be shared

AlessioC31: But in case I think that from the viewer we have the gba/cpu and inside that the InternalMemory is already under `Arc<Mutex<>>` so maybe for now we can remove it from the rom and see if in future we need it

emu/src/memory/internal_memory.rs
AlessioC31: Yup I think it makes sense to add it to this PR since without it we can't read instructions from the ROM

guerinoni: Done

emu/src/memory/internal_memory.rs
guerinoni: Done

emu/src/memory/internal_memory.rs
```diff
@@ -209,19 +214,20 @@ impl InternalMemory {
...
+            0x03000000..=0x03007FFF => self.working_iram[address - 0x03000000],
             0x04000000..=0x04000055 => self.read_address_lcd_register(address),
             0x04000100..=0x0400010E => self.read_address_timer_register(address),
-            0x05000000..=0x050001FF => self.bg_palette_ram[(address - 0x05000000) as usize],
-            0x05000200..=0x050003FF => self.obj_palette_ram[(address - 0x05000200) as usize],
-            0x06000000..=0x06017FFF => self.video_ram[(address - 0x06000000) as usize],
+            0x05000000..=0x050001FF => self.bg_palette_ram[address - 0x05000000],
+            0x05000200..=0x050003FF => self.obj_palette_ram[address - 0x05000200],
+            0x06000000..=0x06017FFF => self.video_ram[address - 0x06000000],
+            0x08000000..=0x0FFFFFFF => self.rom[address - 0x08000000],
```
AlessioC31, line 230: I think this should be splitted in multiple lines. Accordin to gbatek:

> 08000000-09FFFFFF   Game Pak ROM/FlashROM (max 32MB) - Wait State 0
  0A000000-0BFFFFFF   Game Pak ROM/FlashROM (max 32MB) - Wait State 1
  0C000000-0DFFFFFF   Game Pak ROM/FlashROM (max 32MB) - Wait State 2
  0E000000-0E00FFFF   Game Pak SRAM    (max 64 KBytes) - 8bit Bus width
  0E010000-0FFFFFFF   Not used

So I would do something like:

```rust
0x08000000..=0x09FFFFFF => self.rom[address - 0x08000000],
0x0A000000..=0x0BFFFFFF => self.rom[address - 0x0A000000],
0x0C000000..=0x0DFFFFFF => self.rom[address - 0x0C000000],
0x0E000000..=0x0E00FFFF  => self.rom_sram[address - 0x0E000000],
```

And then `rom_sram` should also be implemented in the `write_at` part since it's a ram

AlessioC31 (changes requested):

emu/src/memory/internal_memory.rs
guerinoni: Add a comment for don't forget

AlessioC31 (approved):

PR #130, opened 2022-12-02
Single data transfer fix

AlessioC31:
This should be merged after #129 .

* Fix a bug in `single_data_transfer` function: we were inverting the logic of the up/down bit
* Now cpu boots in supervisor mode (https://www.reddit.com/r/EmuDev/comments/6tr4fz/gba_does_anyone_have_any_information_on_the/ and https://www.cs.rit.edu/~tjh8300/CowBite/CowBiteSpec.htm#SVC). Not 100% sure yet but I think it's still better than booting in User mode :D

### DIFF
--- a/emu/src/arm/arm7tdmi.rs
+++ b/emu/src/arm/arm7tdmi.rs
@@ -32,7 +32,7 @@ impl Default for Arm7tdmi {
     fn default() -> Self {
         let mut s = Self {
             memory: Arc::new(Mutex::new(InternalMemory::default())),
-            cpsr: Psr::from(Mode::User), // FIXME: Starting as User? Not sure
+            cpsr: Psr::from(Mode::Supervisor), // FIXME: Starting as Supervisor? Not sure
             registers: Registers::default(),
             register_bank: RegisterBank::default(),
         };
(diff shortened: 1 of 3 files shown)


emu/src/arm/arm7tdmi.rs
```diff
@@ -32,7 +32,7 @@ impl Default for Arm7tdmi {
     fn default() -> Self {
         let mut s = Self {
             memory: Arc::new(Mutex::new(InternalMemory::default())),
-            cpsr: Psr::from(Mode::User), // FIXME: Starting as User? Not sure
+            cpsr: Psr::from(Mode::Supervisor), // FIXME: Starting as Supervisor? Not sure
```
guerinoni, line 35: Not sure about this, maybe some bios part will call SWI to switch into Supervisor... Default mode is User. I see other reference starting with System mode 🤔

AlessioC31: Uhm I don't know, in many places they say that supervisor is used either at reset state or via SWI.

https://www.linuxbaya.com/2019/08/operating-modes-in-arm-7.html
> Supervisor Mode ARM7 enters this mode on reset. It is used to execute the BIOS program (Booting program). This mode can also be invoked by the programmer by SWI (Software Interrupt). 

https://courses.cs.washington.edu/courses/cse466/10au/pdfs/lectures/07-arm_overview.pdf
> Supervisor : entered on reset and when a Software Interrupt
 instruction is executed

guerinoni: Let's go for `Supervisor` :)

PR #134, opened 2022-12-03
Add interrupts memory region

guerinoni:
- Memory: Refactor lcd as IoDevice
- Memory: Add struct for read/write interrupts area

### DIFF
--- a/emu/src/memory/interrupts.rs
+++ b/emu/src/memory/interrupts.rs
@@ -1,0 +1,65 @@
+use crate::{
+    bitwise::Bits,
+    memory::io_registers::{IORegister, IORegisterAccessControl},
+};
+
+use super::io_device::IoDevice;
+
+pub struct Interrupts {
+    //   4000200h  2    R/W  IE        Interrupt Enable Register
+    //   4000202h  2    R/W  IF        Interrupt Request Flags / IRQ Acknowledge
+    //   4000204h  2    R/W  WAITCNT   Game Pak Waitstate Control
+    //   4000206h       -    -         Not used
+    /// Interrupt Master Enable Register
+    ime: IORegister,
+    //   400020Ah       -    -         Not used
+    /// Post boot flag.
+    post_flag: IORegister,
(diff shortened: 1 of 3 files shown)


emu/src/memory/interrupts.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+impl IoDevice for Interrupts {
+    type Address = usize;
+    type Value = u8;
+
+    fn read_at(&self, address: usize) -> u8 {
+        match address {
+            0x04000300 => self.post_flag.read().get_byte(0),
+            0x04000208 => self.ime.read().get_byte(0),
```
AlessioC31, line 51: `Interrupt Master Enable Register` is two bytes so we should also put `0x04000209 => self.ime.read().get_byte(1),`

emu/src/memory/interrupts.rs
```diff
@@ -0,0 +1,63 @@
...
+            _ => panic!("Reading an write-only memory address: {address:x}"),
+        }
+    }
+
+    fn write_at(&mut self, address: usize, value: u8) {
+        match address {
+            0x04000300 => self.post_flag.set_byte(0, value),
+            0x04000208 => self.ime.set_byte(0, value),
```
AlessioC31, line 59: Same as above

AlessioC31 (changes requested):

emu/src/memory/interrupts.rs
guerinoni: I know, I just want to see if it will go to read that or we need to return a u32 (or u16) instead of u8

AlessioC31: I'd say we should stick with `u8` since the code will probably read `0x04000208` and `0x04000209` subsequentely to get the value of the register

guerinoni: I also think is in this way 🤞

AlessioC31 (approved):

PR #138, opened 2022-12-07
Small fixes in thumb/bx

AlessioC31:
* `CpuState` was using the inverted logic
* `BX` was not reading the value from `Rn`
* Added instruction decode for THUMB mode
* Setting lsb in PC to 0 before fetching
image


guerinoni:
Mostly LGTM, I don't understand if for reading PC we should read 32 bit with first (and in case second) to 0 or if we should read only other 31 (or in case 30)

guerinoni:
Make sense also to me :)

PR #142, opened 2022-12-10
Thumb: Fix the value used to set relative load

guerinoni:

### DIFF
--- a/emu/src/cpu/arm7tdmi.rs
+++ b/emu/src/cpu/arm7tdmi.rs
@@ -611,9 +611,13 @@ impl Arm7tdmi {
 
     pub fn pc_relative_load(&mut self, op_code: ThumbModeOpcode) -> Option<u32> {
         let rd = op_code.get_bits(8..=10);
-        let value = op_code.get_bits(0..=7);
+        let address = op_code.get_bits(0..=7) as usize;
+        let mut pc = self.registers.program_counter() as u32;
+        pc.set_bit_off(1);
+        let address = pc as usize + 4_usize + (address << 2);
+        let value = self.memory.lock().unwrap().read_word(address);
         self.registers
-            .set_register_at(rd.try_into().unwrap(), value as u32);
+            .set_register_at(rd.try_into().unwrap(), value);
 
         Some(SIZE_OF_THUMB_INSTRUCTION)
     }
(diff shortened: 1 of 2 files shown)


guerinoni:
Thanks @AlessioC31 for heads up!

emu/src/cpu/arm7tdmi.rs
```diff
@@ -611,9 +611,12 @@ impl Arm7tdmi {
...
     pub fn pc_relative_load(&mut self, op_code: ThumbModeOpcode) -> Option<u32> {
         let rd = op_code.get_bits(8..=10);
-        let value = op_code.get_bits(0..=7);
+        let address = op_code.get_bits(0..=7) as usize;
+        let value = self.memory.lock().unwrap().read_word(address);
+        let mut pc = self.registers.program_counter() as u32;
+        pc.set_bit_off(0);
         self.registers
-            .set_register_at(rd.try_into().unwrap(), value as u32);
+            .set_register_at(rd.try_into().unwrap(), (pc + 4 + value) << 2);
```
AlessioC31, line 619: I think the address should be `pc + 4 + (immediate << 2)` because we should read an address which is relative to the PC. What do you think?

AlessioC31 (changes requested):

emu/src/cpu/arm7tdmi.rs
guerinoni: Yes seems you're right

AlessioC31: Mhh I still think there's something off. What I understood from the documentation is that we construct the address by doing `pc + 4 + (value << 2)` and then we read the memory at that address. Right now we're reading at `imm8` address.

AlessioC31: Something like:

```rust
let address = pc + 4 + (value << 2); // where value is the immediate in the opcode
let value = self.memory.lock().unwrap().read_word(address);
self.registers.set_register_at(rd, address);
```

guerinoni: Ok now looks correct :)

emu/src/cpu/arm7tdmi.rs
```diff
@@ -611,9 +611,13 @@ impl Arm7tdmi {
 
     pub fn pc_relative_load(&mut self, op_code: ThumbModeOpcode) -> Option<u32> {
         let rd = op_code.get_bits(8..=10);
-        let value = op_code.get_bits(0..=7);
+        let address = op_code.get_bits(0..=7) as usize;
+        let mut pc = self.registers.program_counter() as u32;
+        pc.set_bit_off(0);
```
AlessioC31, line 616:
```suggestion
        pc.set_bit_off(1);
```

Last one, sorry! In THUMB we're sure we've bit 0 to 0 since we set it before fetching, we have to set bit 1 to 0 here.

guerinoni: Oh yes because only in Arm mode we set both to zero, right

AlessioC31 (approved):

PR #143, opened 2022-12-12
THUMB: Implement `Load/Store Register offset` instruction

AlessioC31:
* Add `write_word` function to internal memory
* Implement instruction
* Rewrote `get_bits` in order to avoid using `for` and using binary arithmetics instead. It's faster than 50% on `u32` (but this 50% is only 20ns, lol).

Should we move `ReadWriteKind` somewhere else since it's not used only by `single_data_transfer`?

### DIFF
--- a/emu/src/cpu/arm7tdmi.rs
+++ b/emu/src/cpu/arm7tdmi.rs
@@ -15,10 +15,25 @@ use crate::memory::io_device::IoDevice;
 use super::opcode::ThumbModeOpcode;
 use super::psr::CpuState;
 use super::registers::Registers;
+use super::single_data_transfer::ReadWriteKind;
 
 pub const REG_PROGRAM_COUNTER: u32 = 0xF;
 pub const SIZE_OF_ARM_INSTRUCTION: u32 = 4;
 pub const SIZE_OF_THUMB_INSTRUCTION: u32 = 2;
+
+enum LoadStoreKind {
+    Store,
+    Load,
+}
+
+impl From<bool> for LoadStoreKind {
+    fn from(b: bool) -> Self {
+        match b {
+            false => Self::Store,
+            true => Self::Load,
+        }
+    }
+}
 
 pub struct Arm7tdmi {
     pub(crate) memory: Arc<Mutex<InternalMemory>>,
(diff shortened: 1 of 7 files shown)


emu/src/memory/internal_memory.rs
```diff
@@ -135,6 +135,18 @@ impl InternalMemory {
         part_3 << 24_u32 | part_2 << 16_u32 | part_1 << 8_u32 | part_0
     }
 
+    pub fn write_word(&mut self, address: usize, value: u32) {
+        let part_0: u8 = (value & 0xFF).try_into().unwrap();
+        let part_1: u8 = ((value & (0xFF << 8)) >> 8).try_into().unwrap();
+        let part_2: u8 = ((value & (0xFF << 16)) >> 16).try_into().unwrap();
+        let part_3: u8 = ((value & (0xFF << 24)) >> 24).try_into().unwrap();
```
guerinoni, line 142: For the sake of fun, can be better in terms of readability write something like
`value.get_bits(0..=3)`
`value.get_bits(4..=7)` ...

emu/src/cpu/arm7tdmi.rs
```diff
@@ -224,6 +238,43 @@ impl Arm7tdmi {
...
+
+        let rd: usize = op_code.get_bits(0..=2).try_into().unwrap();
+
+        let address: usize = rb.wrapping_add(ro).try_into().unwrap();
+
+        let mut mem = self.memory.lock().unwrap();
+
+        match (load_store, byte_word) {
+            (LoadStoreKind::Store, ReadWriteKind::Byte) => {
+                let rd = (self.registers.register_at(rd) & 0xFF) as u8;
```
guerinoni, line 258: `self.registers.register_at(rd)` returns a u32, why `& 0xFF` and cast as u8? You are not considering 8 bits here

emu/src/cpu/single_data_transfer.rs
```diff
@@ -32,7 +32,7 @@ impl From<ArmModeOpcode> for SingleDataTransfer {
 
 /// There two different kind of write or read for memory.
 #[derive(Default)]
-enum ReadWriteKind {
+pub enum ReadWriteKind {
```
guerinoni, line 35: Yes we can create a module of `OperationsFlag`? Like pre/post load/store and so on...

emu/src/memory/internal_memory.rs
AlessioC31: I know but `get_bits` uses a for which I don't like in this case since it's slower. Maybe I can rewrite `get_bits` and then I can use it here

emu/src/cpu/single_data_transfer.rs
AlessioC31: Done!

emu/src/memory/internal_memory.rs
AlessioC31: Done

guerinoni (approved):
Nice :)

PR #145, opened 2022-12-13
Thumb: Implement BX Hs (only the specific case)

guerinoni:

### DIFF
--- a/emu/src/cpu/arm7tdmi.rs
+++ b/emu/src/cpu/arm7tdmi.rs
@@ -101,8 +101,6 @@ impl Arm7tdmi {
             .advance_program_counter(bytes_to_advance.unwrap_or(0));
     }
 
-    #[allow(unreachable_code)]
-    #[allow(unused_variables)]
     pub fn execute_thumb(&mut self, op_code: ThumbModeOpcode) {
         use ThumbModeInstruction::*;
         let bytes_to_advance: Option<u32> = match op_code.instruction {
@@ -110,7 +108,7 @@ impl Arm7tdmi {
             AddSubtract => self.add_subtract(op_code),
             MoveCompareAddSubtractImm => self.move_compare_add_sub_imm(op_code),
             AluOp => unimplemented!(),
-            HiRegisterOpBX => unimplemented!(),
+            HiRegisterOpBX => self.hi_reg_operation_branch_ex(op_code),
             PCRelativeLoad => self.pc_relative_load(op_code),
             LoadStoreRegisterOffset => self.load_store_register_offset(op_code),
             LoadStoreSignExtByteHalfword => unimplemented!(),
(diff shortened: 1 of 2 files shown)


emu/src/cpu/arm7tdmi.rs
```diff
@@ -680,6 +678,48 @@ impl Arm7tdmi {
...
     }
+
+    pub(crate) fn hi_reg_operation_branch_ex(&mut self, op_code: ThumbModeOpcode) -> Option<u32> {
+        let op = op_code.get_bits(8..=9);
+        let h1 = op_code.get_bit(7);
+        let h2 = op_code.get_bit(6);
+        let rs_hs = op_code.get_bits(3..=5);
+        let _rd_hd = op_code.get_bits(0..=2);
+
+        match (op.get_bit(1), op.get_bit(0)) {
```
AlessioC31, line 689: I would to the match on the value:

```rust
// Add
0b00 => {},
// Cmp
0b01 => {},
// Mov
0b10 => {},
// Bx
0b11 => {},
_ => unreachable!()

File: emu/src/cpu/arm7tdmi.rs [hunk 2]
```diff
@@ -680,6 +678,48 @@ impl Arm7tdmi {
...
         Some(SIZE_OF_THUMB_INSTRUCTION)
     }
+
+    pub(crate) fn hi_reg_operation_branch_ex(&mut self, op_code: ThumbModeOpcode) -> Option<u32> {
+        let op = op_code.get_bits(8..=9);
+        let h1 = op_code.get_bit(7);
+        let h2 = op_code.get_bit(6);
+        let rs_hs = op_code.get_bits(3..=5);
+        let _rd_hd = op_code.get_bits(0..=2);
+
```
AlessioC31, line 688: We could do something like:

```rust
let destination_register: usize = (h1 as u16 * 8 + rd_hd).try_into().unwrap();
let source_register: usize = (h2 as u16 * 8 + rs_hs).try_into().unwrap();
```

So we avoid the ifs later

emu/src/cpu/arm7tdmi.rs
```diff
@@ -680,6 +678,48 @@ impl Arm7tdmi {
...
+                let v = self.registers.register_at(rs_hs.try_into().unwrap());
+                if !v.get_bit(0) {
+                    self.cpsr.set_cpu_state(CpuState::Arm);
+                } else {
+                    self.cpsr.set_cpu_state(CpuState::Thumb);
+                }
+                if rs_hs as u32 == REG_PROGRAM_COUNTER {
+                    todo!()
+                }
+                let v = v & 0xFFFF_FFFE;
```
AlessioC31, line 713: I think we don't need it here since we word-align/halfword-align in the `step` function of the cpu

emu/src/cpu/arm7tdmi.rs
```diff
@@ -680,6 +678,48 @@ impl Arm7tdmi {
...
+                self.registers.set_program_counter(v);
+            }
+            (true, true) if h1 && !h2 => todo!(),
+            (true, true) if h1 && h2 => todo!(),
+            (true, true) => unreachable!(),
+        };
+
+        Some(SIZE_OF_THUMB_INSTRUCTION)
```
AlessioC31, line 721: If we branch we should return `None`

AlessioC31 (changes requested):

emu/src/cpu/arm7tdmi.rs
guerinoni: Nice idea!

emu/src/cpu/arm7tdmi.rs
guerinoni: Done

emu/src/cpu/arm7tdmi.rs
guerinoni: Removed :)

AlessioC31 (approved):

PR #146, opened 2022-12-13
THUMB: Complete `HiRegOpBX` instruction

AlessioC31:


emu/src/cpu/arm7tdmi.rs
```diff
@@ -686,29 +686,63 @@ impl Arm7tdmi {
...
+                REG_PROGRAM_COUNTER => 4,
+                _ => 0,
+            };
+
         match op {
             // Add
-            0b00 => todo!(), //Some(SIZE_OF_THUMB_INSTRUCTION)
+            0b00 => {
+                self.registers
+                    .set_register_at(destination_register, first_op + second_op);
```
guerinoni, line 714: Simple `+` or `add_inner`?

guerinoni (approved):
Nice tests

emu/src/cpu/arm7tdmi.rs
AlessioC31: Simple `+` is ok since only MOV changes CPSR in this instruction

PR #147, opened 2022-12-14
Register Bank fix. Fix MSR op

AlessioC31:
Merge after #146 

* Fix mode swapping and register bank logic
Before, we were using the different SPSR_{mode} as saved versions
of CPSR. This is wrong, the correct way of doing is that every mode
(except System and User) have their SPSR, separate from the CPSR.
So, the SPSR is just another register (analogous to the CPSR) that gets
banked for every mode.
Another problem was also that we were not saving/restoring the "base"
r8-14 registers. So for example when going from System to FIQ, FIQ
was replacing r8-14 with its own versions. Then, if the system would
have gone from FIQ to System the old values of r8-14 would have not
been restored.

* ARM: Fix MSR op
Now it correctly stores/restores banked registers.

### DIFF
--- a/emu/src/cpu/arm7tdmi.rs
+++ b/emu/src/cpu/arm7tdmi.rs
@@ -25,6 +25,7 @@ pub struct Arm7tdmi {
     pub(crate) memory: Arc<Mutex<InternalMemory>>,
 
     pub cpsr: Psr,
+    pub spsr: Psr,
     pub registers: Registers,
 
     pub register_bank: RegisterBank,
@@ -35,6 +36,7 @@ impl Default for Arm7tdmi {
         let mut s = Self {
             memory: Arc::new(Mutex::new(InternalMemory::default())),
             cpsr: Psr::from(Mode::Supervisor), // FIXME: Starting as Supervisor? Not sure
+            spsr: Psr::default(),
             registers: Registers::default(),
             register_bank: RegisterBank::default(),
         };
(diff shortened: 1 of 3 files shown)


emu/src/cpu/arm7tdmi.rs
```diff
@@ -421,11 +412,14 @@ impl Arm7tdmi {
...
-    /// It should be used before leaving the current mode.
-    pub fn store_registers_in_bank(&mut self) {
+    pub fn swap_mode(&mut self, new_mode: Mode) {
+        if self.cpsr.mode() == new_mode {
+            return;
+        }
+
         match self.cpsr.mode() {
-            Mode::User | Mode::System => {}
+            // If we leave Fiq we store r8-14 and spsr.
```
guerinoni, line 421: Maybe are you missing to store `spsr`?

emu/src/cpu/data_processing.rs
```diff
@@ -276,27 +276,41 @@ impl Arm7tdmi {
...
+                {
+                    let psr = match p {
+                        false => &mut self.cpsr,
+                        true => &mut self.spsr,
+                    };
+                    // Setting flags
+                    psr.set_sign_flag(rm.get_bit(31));
+                    psr.set_zero_flag(rm.get_bit(30));
+                    psr.set_carry_flag(rm.get_bit(29));
+                    psr.set_overflow_flag(rm.get_bit(28));
```
guerinoni, line 292: You can use `pub fn set_flags(&mut self, op_result: ArithmeticOpResult)` for this purpose

emu/src/cpu/data_processing.rs
```diff
@@ -276,27 +276,41 @@ impl Arm7tdmi {
...
+
+                    // In User mode we can only set the flags so we don't touch the Mode bits
+                    if current_mode != Mode::User {
+                        psr.set_irq_disable(rm.get_bit(7));
+                        psr.set_fiq_disable(rm.get_bit(6));
+
+                        // Documentation says that software should never touch T (state) bit
+                        // Should we set it? I guess software are written in order to not switch this bit
+                        // but who knows?
+                        // psr.set_state_bit(rm.get_bit(5));
```
guerinoni, line 302: Big question 😂 I would try to keep as is and what is going on

emu/src/cpu/data_processing.rs
AlessioC31: But then we shold construct a "dummy" `ArithmeticOpResult`?

emu/src/cpu/arm7tdmi.rs
AlessioC31: Yes indeed, thank you!

emu/src/cpu/data_processing.rs
guerinoni: Yes is a borderline case

guerinoni (approved):
LGTM

emu/src/cpu/data_processing.rs
AlessioC31: But I don't know it would look weird to have an `ArithmeticOpResult` here without any arithmetic operation 😅

PR #148, opened 2022-12-15
Partially implementation of Push/Pop

guerinoni:
- Thumb: Implement partially Push/Pop
- Everwhere: Fix lint after bump rust to 1.66

### DIFF
--- a/emu/src/cpu/arm7tdmi.rs
+++ b/emu/src/cpu/arm7tdmi.rs
@@ -17,6 +17,8 @@ use super::opcode::ThumbModeOpcode;
 use super::psr::CpuState;
 use super::registers::Registers;
 
+pub const REG_SP: usize = 0xD;
+pub const REG_LR: usize = 0xE;
 pub const REG_PROGRAM_COUNTER: u32 = 0xF;
 pub const SIZE_OF_ARM_INSTRUCTION: u32 = 4;
 pub const SIZE_OF_THUMB_INSTRUCTION: u32 = 2;
@@ -119,7 +121,7 @@ impl Arm7tdmi {
             SPRelativeLoadStore => unimplemented!(),
             LoadAddress => unimplemented!(),
             AddOffsetSP => unimplemented!(),
-            PushPopReg => unimplemented!(),
+            PushPopReg => self.push_pop_register(op_code),
             MultipleLoadStore => unimplemented!(),
             CondBranch => self.cond_branch(op_code),
             Swi => unimplemented!(),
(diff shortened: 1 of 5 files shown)


emu/src/cpu/arm7tdmi.rs
```diff
@@ -770,6 +772,42 @@ impl Arm7tdmi {
...
+                    if rlist.get_bit(r) {
+                        let reg_sp = self.registers.register_at(REG_SP);
+                        self.registers.set_register_at(REG_SP, reg_sp - 4);
+                        let stack_addr = (self.registers.register_at(REG_SP) & !3) as usize;
+                        self.memory
+                            .lock()
+                            .unwrap()
+                            .write_word(stack_addr, self.registers.register_at(r.into()));
+                    }
+                }
```
AlessioC31, line 802: LGTM in general, I would only rewrite it in a way where we don't read/store every time the sp but we read it once at the beginning, we use it, we decrement it each time and then only at the end we write it back.

Also, I'm not sure about the `& !3`, I know it's in order to have a word-aligned address but documentation says that the 2 lsb might be interpreted by the memory system (maybe we will need to do it in the future?). And also in the implementation of the ldm/stm we are not doing `& !3`. Maybe we can avoid doing it and we put a warning in the `write_word`/`read_word` in the `InternalMemory` if the provided address is not word-aligned? So we know if something is off or if we should interpret them somehow

emu/src/cpu/arm7tdmi.rs
```diff
@@ -770,6 +772,42 @@ impl Arm7tdmi {
...
+        match load_store {
+            LoadStoreKind::Store if !store_or_not_reg_lr => todo!(),
+            LoadStoreKind::Store if store_or_not_reg_lr => {
+                let reg_sp = self.registers.register_at(REG_SP);
+                self.registers.set_register_at(REG_SP, reg_sp - 4);
+                let stack_addr = (self.registers.register_at(REG_SP) & !3) as usize;
+                self.memory
+                    .lock()
+                    .unwrap()
+                    .write_word(stack_addr, self.registers.register_at(REG_LR));
```
AlessioC31, line 790: Also instead of splitting it in two match arms, what do you whink if we do the if inside the `LoadStoreKind::Store` arm?

Something like that:

```rust
LoadStoreKind::Store => {
    if store_or_not_reg_lr {
        let reg_sp = self.registers.register_at(REG_SP);
        self.registers.set_register_at(REG_SP, reg_sp - 4);
        let stack_addr = (self.registers.register_at(REG_SP) & !3) as usize;
        self.memory
            .lock()
            .unwrap()
            .write_word(stack_addr, self.registers.register_at(REG_LR));
    }
    ...
}, ...
```

AlessioC31 (changes requested):

emu/src/cpu/arm7tdmi.rs
guerinoni: Improved

emu/src/cpu/arm7tdmi.rs
guerinoni: Right now I removed! I will add a little warning in r/w word

AlessioC31 (approved):

PR #152, opened 2022-12-16
Alu operations

guerinoni:
- Thumb: Add alu operation
- Thumb: Implement `mvn` for ALU

### DIFF
--- a/emu/src/cpu/arm7tdmi.rs
+++ b/emu/src/cpu/arm7tdmi.rs
@@ -13,6 +13,7 @@ use crate::cpu::register_bank::RegisterBank;
 use crate::memory::internal_memory::InternalMemory;
 use crate::memory::io_device::IoDevice;
 
+use super::alu_instruction::ThumbModeAluInstruction;
 use super::flags::{Indexing, LoadStoreKind, Offsetting, OperandKind, ReadWriteKind};
 use super::opcode::ThumbModeOpcode;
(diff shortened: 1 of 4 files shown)


emu/src/cpu/arm7tdmi.rs
```diff
@@ -873,6 +874,35 @@ impl Arm7tdmi {
...
+            ThumbModeAluInstruction::Ror => todo!(),
+            ThumbModeAluInstruction::Tst => todo!(),
+            ThumbModeAluInstruction::Neg => todo!(),
+            ThumbModeAluInstruction::Cmp => todo!(),
+            ThumbModeAluInstruction::Cmn => todo!(),
+            ThumbModeAluInstruction::Orr => todo!(),
+            ThumbModeAluInstruction::Mul => todo!(),
+            ThumbModeAluInstruction::Bic => todo!(),
+            ThumbModeAluInstruction::Mvn => {
+                self.mvn(rd.try_into().unwrap(), rs.try_into().unwrap(), false);
```
AlessioC31, line 900: I think the last parameter should be `true` since these alu ops set the flags

AlessioC31 (approved):

AlessioC31 (commented):
I think the last parameter should be `true` since these alu ops set the flags. 

I didn't click on the line but I mean when you call `self.mvn`

AlessioC31 (approved):

PR #153, opened 2022-12-16
Implement Long Branch Link

guerinoni:
- Thumb: Add alu operation
- Thumb: Implement `mvn` for ALU
- Thumb: Implement `Long Branch Link`

### DIFF
--- a/emu/src/cpu/arm7tdmi.rs
+++ b/emu/src/cpu/arm7tdmi.rs
@@ -128,7 +128,7 @@ impl Arm7tdmi {
             CondBranch => self.cond_branch(op_code),
             Swi => unimplemented!(),
             UncondBranch => unimplemented!(),
-            LongBranchLink => unimplemented!(),
+            LongBranchLink => self.long_branch_link(op_code),
         };
 
         self.registers
@@ -903,6 +903,29 @@ impl Arm7tdmi {
 
         Some(SIZE_OF_THUMB_INSTRUCTION)
     }
+
+    fn long_branch_link(&mut self, op_code: ThumbModeOpcode) -> Option<u32> {
+        let h = op_code.get_bit(11);
+        let offset = op_code.get_bits(0..=10);
+
+        if h {
+            let next_instruction =
+                self.registers.program_counter() as u32 + SIZE_OF_THUMB_INSTRUCTION;
+            let lr = self.registers.register_at(REG_LR);
+            let offset = (offset << 1) as u32;
+            self.registers.set_program_counter(lr.wrapping_add(offset));
+            self.registers.set_register_at(REG_LR, next_instruction | 1);
+        } else {
+            let offset = (offset << 12) as u32;
+            let pc = self.registers.program_counter() as u32
+                + SIZE_OF_THUMB_INSTRUCTION
+                + SIZE_OF_THUMB_INSTRUCTION;
+            self.registers
+                .set_register_at(REG_LR, pc.wrapping_add(offset));
+        }
+
+        Some(SIZE_OF_THUMB_INSTRUCTION)
+    }
 }
 
 pub enum HalfwordTransferType {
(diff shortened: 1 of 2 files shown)


emu/src/cpu/arm7tdmi.rs
```diff
@@ -903,6 +903,26 @@ impl Arm7tdmi {
...
+                self.registers.program_counter() as u32 + SIZE_OF_THUMB_INSTRUCTION;
+            let lr = self.registers.register_at(REG_LR);
+            let offset = (offset << 1) as u32;
+            self.registers.set_program_counter(lr + offset);
+            self.registers.set_register_at(REG_LR, next_instruction | 1);
+        } else {
+            let offset = (offset << 12) as u32;
+            let pc = self.registers.program_counter() as u32;
+            self.registers.set_register_at(REG_LR, offset + pc);
+        }
```
AlessioC31, line 922: I think we should use `.wrapping_add`.

Imagine this example, which for semplicity is a 4bits example:

`0b0100 + 0b1110` (`4 - 2`). If we only do `+` we have a panic since it overflows but it's ok to overflow since then we would have `0b0010` which is the correct result

emu/src/cpu/arm7tdmi.rs
```diff
@@ -903,6 +903,26 @@ impl Arm7tdmi {
...
+            let next_instruction =
+                self.registers.program_counter() as u32 + SIZE_OF_THUMB_INSTRUCTION;
+            let lr = self.registers.register_at(REG_LR);
+            let offset = (offset << 1) as u32;
+            self.registers.set_program_counter(lr + offset);
+            self.registers.set_register_at(REG_LR, next_instruction | 1);
+        } else {
+            let offset = (offset << 12) as u32;
+            let pc = self.registers.program_counter() as u32;
+            self.registers.set_register_at(REG_LR, offset + pc);
```
AlessioC31, line 921: `+4` for prefetching

emu/src/cpu/arm7tdmi.rs
```diff
@@ -903,6 +903,26 @@ impl Arm7tdmi {
...
+        let h = op_code.get_bit(11);
+        let offset = op_code.get_bits(0..=10);
+
+        if h {
+            let next_instruction =
+                self.registers.program_counter() as u32 + SIZE_OF_THUMB_INSTRUCTION;
+            let lr = self.registers.register_at(REG_LR);
+            let offset = (offset << 1) as u32;
+            self.registers.set_program_counter(lr + offset);
+            self.registers.set_register_at(REG_LR, next_instruction | 1);
```
AlessioC31, line 917: Did you understand why we set bit 0? I can't figure it out why

AlessioC31: Is it because if we later do a `bx lr` then we take the bit 0 as `1` and it means we would still use THUMB?

emu/src/cpu/arm7tdmi.rs
guerinoni: Right!!

emu/src/cpu/arm7tdmi.rs
guerinoni: Maybe +2 as we are in thumb mode

emu/src/cpu/arm7tdmi.rs
guerinoni: Don't remember right now, I need to recall my walkthrough ehehe but maybe yes!

emu/src/cpu/arm7tdmi.rs
AlessioC31: But we prefetch 2 istructions

guerinoni: Mmmm ok fixed

AlessioC31 (approved):

PR #154, opened 2022-12-20
Thumb: Implement LSL for shift

guerinoni:
- CPU: Create `ShifKind` for u32/u16 conversion
- CPU: Move `ArithmeticOpResult` into alu_instruction module
- Thumb: Implement `LSL` for shift register

### DIFF
--- a/emu/src/cpu/alu_instruction.rs
+++ b/emu/src/cpu/alu_instruction.rs
@@ -138,6 +138,15 @@ impl From<u32> for ShiftKind {
     }
 }
 
+#[derive(Default)]
+pub struct ArithmeticOpResult {
+    pub result: u32,
+    pub carry: bool,
+    pub overflow: bool,
+    pub sign: bool,
+    pub zero: bool,
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/emu/src/cpu/alu_instruction.rs
+++ b/emu/src/cpu/alu_instruction.rs
@@ -1,3 +1,5 @@
+use crate::bitwise::Bits;
+
 pub enum ArmModeAluInstruction {
     And = 0x0,
     Eor = 0x1,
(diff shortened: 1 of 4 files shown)


emu/src/cpu/alu_instruction.rs
```diff
@@ -147,6 +149,34 @@ pub struct ArithmeticOpResult {
     pub zero: bool,
 }
 
+pub fn shitf(kind: ShiftKind, shift_amount: u32, rm: u32, _carry: bool) -> ArithmeticOpResult {
```
AlessioC31, line 152: `shift`?

AlessioC31 (changes requested):
Since we moved the `shift` logic to a new function, we lose the fact that before the code was modifying the `carry` variable and later changing it in the cpu. Now the `shift` function doesn't change the carry, thus the `shift_operand` function doesn't updates the `carry` bit in the cpsr.

I'm not sure about using `ArithmeticOpResult` here since it contains fields we would never use in such a context. Maybe we can rethink it?

emu/src/cpu/alu_instruction.rs
guerinoni: Lol

guerinoni:
> Since we moved the `shift` logic to a new function, we lose the fact that before the code was modifying the `carry` variable and later changing it in the cpu. Now the `shift` function doesn't change the carry, thus the `shift_operand` function doesn't updates the `carry` bit in the cpsr, since it only uses the `result` field of the returned value from `shift`
> 
> I'm not sure about using `ArithmeticOpResult` here since it contains fields we would never use in such a context. The idea behind this struct was to carry the "new state" of an ALU operation (add/sub). Maybe we can rethink it? Maybe we could make it more "modular"

Ok now I re-designed a little bit and seems correct to me! Obviously the refactor will be nicer when all operation will be split as I did for `shift`... But meanwhile... :)

AlessioC31 (commented):
I still think tha `ArithmeticOpResult` doesn't fit here 100% but we can first merge this and maybe in another PR we can rethink this entire thing

emu/src/cpu/alu_instruction.rs
```diff
@@ -147,6 +149,34 @@ pub struct ArithmeticOpResult {
...
 }
 
+pub fn shift(kind: ShiftKind, shift_amount: u32, rm: u32, _carry: bool) -> ArithmeticOpResult {
+    match kind {
+        ShiftKind::Lsl => {
+            match shift_amount {
+                // LSL#0: No shift performed, ie. directly value=Rm, the C flag is NOT affected.
+                0 => ArithmeticOpResult {
+                    result: rm,
+                    ..Default::default()
```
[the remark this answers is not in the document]
guerinoni: Oh yes! Nice catch

AlessioC31 (approved):

Issue #32, opened 2022-09-07
Returned CPSR flags must be implemented for ALU instructions


gallottino:
Reference [here](https://www.problemkaputt.de/gbatek.htm#armopcodesdataprocessingalu).

AlessioC31:
Implemented

PR #155, opened 2022-12-22
Add UI for disassembler

guerinoni:
This PR prepare the code for supporting the ASM widget.
Now half of work we previous did in same function of execution a specific instruction is moved while decoding, this because we can use those information for printing all parameters in a widgets.

Also the tests are very complete right now because we can test many different condition in a clear way.

https://user-images.githubusercontent.com/41150432/218267824-02f16fcb-7af7-4c6f-8916-9864ee409280.mov


emu/src/cpu/instruction.rs
```diff
@@ -4,34 +4,66 @@ use logger::log;
 
 use crate::bitwise::Bits;
 
+use super::condition::Condition;
+
 #[derive(Debug, PartialEq, Eq)]
 pub enum ArmModeInstruction {
```
AlessioC31, line 10: Is there the possiblity to have `Cond` for every option without the need to specify it everytime?

emu/src/cpu/instruction.rs
```diff
@@ -4,34 +4,66 @@ use logger::log;
...
+            Self::Multiply => "".to_owned(),
+            Self::MultiplyLong => "".to_owned(),
+            Self::SingleDataSwap => "".to_owned(),
+            Self::BranchAndExchange(_, reg) => format!("bx R{}", reg),
+            Self::HalfwordDataTransferRegisterOffset => "".to_owned(),
+            Self::HalfwordDataTransferImmediateOffset => "".to_owned(),
+            Self::SingleDataTransfer => "".to_owned(),
+            Self::Undefined => "".to_owned(),
+            Self::BlockDataTransfer => "".to_owned(),
+            Self::Branch(cond, is_link, offset) => {
```
AlessioC31, line 40: The assembler syntax for branch/branchlink is `B{L}{cond} <expression>` where expression is the address where we want to branch to, not the offset. The offset is then encoded in the opcode but the asm syntax shows the final address

AlessioC31 (changes requested):
First round of review :D

emu/src/cpu/instruction.rs
guerinoni: Do you mean for every variant of enum? Nope

guerinoni: Or with a macro but...

emu/src/cpu/instruction.rs
guerinoni: Ok, I can't do this without changing many things... Because for having address I need to access to `PC`

AlessioC31: Uhm I get it but if we want to do a disassembler I think we should stick to the official format

emu/src/cpu/instruction.rs
```diff
@@ -3,35 +3,113 @@ use std::fmt::{Display, Formatter};
...
+                    | ArmModeAluInstruction::Bic => {
+                        format!(
+                            "{alu_instruction}{set_string}{condition} R{destination}, 0x{op1:08X}, 0x{op2:08X}"
+                        )
+                    }
+                    ArmModeAluInstruction::Tst
+                    | ArmModeAluInstruction::Teq
+                    | ArmModeAluInstruction::Cmp
+                    | ArmModeAluInstruction::Cmn => {
+                        format!("{alu_instruction}{set_string}{condition} 0x{op1:08X}, 0x{op2:08X}")
```
AlessioC31, line 69: There's not `s` in the ASM syntax for these instructions since they always have the bit `S` set.

And I think we should not display `op1` but `Rn` instead

emu/src/cpu/instruction.rs
```diff
@@ -3,35 +3,113 @@ use std::fmt::{Display, Formatter};
...
+                    }
+                    ArmModeAluInstruction::Mov | ArmModeAluInstruction::Mvn => {
+                        format!(
+                            "{alu_instruction}{set_string}{condition} R{destination}, 0x{op2:08X}"
```
AlessioC31, line 73: `set_string` and `condition` should be inverted

emu/src/cpu/instruction.rs
```diff
@@ -3,35 +3,113 @@ use std::fmt::{Display, Formatter};
...
+                    | ArmModeAluInstruction::Sub
+                    | ArmModeAluInstruction::Rsb
+                    | ArmModeAluInstruction::Add
+                    | ArmModeAluInstruction::Adc
+                    | ArmModeAluInstruction::Sbc
+                    | ArmModeAluInstruction::Rsc
+                    | ArmModeAluInstruction::Orr
+                    | ArmModeAluInstruction::Bic => {
+                        format!(
+                            "{alu_instruction}{set_string}{condition} R{destination}, 0x{op1:08X}, 0x{op2:08X}"
```
AlessioC31, line 62: `set_string` and `cond` inverted here as well.

And I think also here we should display `Rn` and not `op1`

emu/src/cpu/instruction.rs
```diff
@@ -3,35 +3,113 @@ use std::fmt::{Display, Formatter};
...
+                        format!(
+                            "{alu_instruction}{set_string}{condition} R{destination}, 0x{op2:08X}"
+                        )
+                    }
+                }
+            }
+            Self::Multiply => "".to_owned(),
+            Self::MultiplyLong => "".to_owned(),
+            Self::SingleDataSwap => "".to_owned(),
+            Self::BranchAndExchange(_, reg) => format!("bx R{reg}"),
```
AlessioC31, line 81: We miss the cond in the string here

emu/src/cpu/instruction.rs
```diff
@@ -3,35 +3,113 @@ use std::fmt::{Display, Formatter};
...
+            Self::HalfwordDataTransferRegisterOffset => "".to_owned(),
+            Self::HalfwordDataTransferImmediateOffset => "".to_owned(),
+            Self::SingleDataTransfer => "".to_owned(),
+            Self::Undefined => "".to_owned(),
+            Self::BlockDataTransfer => "".to_owned(),
+            Self::Branch(cond, is_link, address) => {
+                let link = if *is_link { "l" } else { "" };
+                format!("b{cond}{link} 0x{address:08X}")
```
AlessioC31, line 89: `cond` and `link` are inverted

AlessioC31 (changes requested):
That's super cool but there are still some things which are not compliant to the standard syntax 😅

From documentation for data processing assembler syntax:

image

So we shouldn't directly show `op2` but we should show `Rm{, <shift>}` if the second operand was a register, or `<#expression>` if it was an immediate

emu/src/cpu/instruction.rs
```diff
@@ -58,20 +58,16 @@ impl ArmModeInstruction {
                     | ArmModeAluInstruction::Rsc
                     | ArmModeAluInstruction::Orr
                     | ArmModeAluInstruction::Bic => {
-                        format!(
-                            "{alu_instruction}{set_string}{condition} R{destination}, 0x{op1:08X}, 0x{op2:08X}"
-                        )
+                        format!("{alu_instruction}{condition}{condition} R{destination}, R{rn}")
```
AlessioC31, line 61: `condition` two times :D

emu/src/cpu/instruction.rs
```diff
@@ -67,11 +187,77 @@ impl From<u32> for ArmModeInstruction {
...
+            let byte_or_word: ReadWriteKind = op_code.into(); // TODO: is this the same for all instruction?
+            let load_store: SingleDataTransferKind = op_code.into(); // TODO: is this the same bit for all instruction?
+            let write_back = op_code.get_bit(21);
+            let rn = op_code.get_bits(16..=19);
+            let rd = op_code.get_bits(12..=15);
+
+            let offset = match op_kind {
+                OperandKind::Immediate => todo!(),
+                OperandKind::Register => op_code.get_bits(0..=11),
+            };
```
AlessioC31, line 206: I think that we have to move this logic in the execute part because we can't read the register here in the decode.

Also, the logic here is inverted, if it is immediate we read `0..=11` and we use it as offset

emu/src/cpu/instruction.rs
```diff
@@ -67,11 +187,77 @@ impl From<u32> for ArmModeInstruction {
         } else if op_code.get_bits(25..=27) == 0b100 {
             BlockDataTransfer
         } else if op_code.get_bits(25..=27) == 0b101 {
-            Branch
+            let is_link = op_code.get_bit(24);
+            let offset = op_code.get_bits(0..=23) << 2;
+            Branch(condition, is_link, offset)
         } else if op_code.get_bits(26..=27) == 0b01 {
-            SingleDataTransfer
+            let op_kind: OperandKind = op_code.get_bit(25).into();
```
AlessioC31, line 194: Thoughts for the future: maybe we should find a better name for `OperandKind` if we want to use it everywhere, because for example here it is not really an operand but the offset, but ok I guess that for now we can leave it

emu/src/cpu/data_processing.rs
```diff
@@ -24,10 +25,55 @@ enum PsrOpKind {
...
+            Self::Immediate(value) => write!(f, "#{value}"),
+            Self::Register(register) => write!(f, "{register}"),
+        }
+    }
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum AluSecondOperandInfo {
+    Register {
+        op_kind: ShiftOperator,
```
AlessioC31, line 46: Rename the parameter in `shift_op`?

emu/src/cpu/arm7tdmi.rs
```diff
@@ -1023,18 +1057,50 @@ impl From<u8> for HalfwordTransferType {
...
-    fn test_branch() {
+    fn check_branch_and_exchange() {
+        {
+            let op_code = 0b1110_0_0_0_1_0_0_1_0_1_1_1_1_1_1_1_1_1_1_1_1_0_0_0_1_0000;
+            let cpu = Arm7tdmi::default();
+            let op_code: ArmModeOpcode = cpu.decode(op_code);
+            assert_eq!(op_code.instruction, BranchAndExchange(Condition::AL, 0));
+            let asm = op_code.instruction.disassembler();
+            assert_eq!(asm, "BX R0");
+        }
```
AlessioC31, line 1074: Here we're not really testing for the functionality, right?

emu/src/cpu/data_processing.rs
```diff
@@ -24,10 +25,55 @@ enum PsrOpKind {
...
+
+impl std::fmt::Display for AluSecondOperandInfo {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match *self {
+            Self::Register {
+                op_kind,
+                shift_kind,
+                register,
+            } => {
+                write!(f, "{register}, {shift_kind} {op_kind}")
```
AlessioC31, line 64: Here we miss a bit of logic in the case of immediate-specified shift, if we do a `LSL #0` it means that we don't do any shift so we don't show anything. `LSR #0` and `ASR #0` they mean `LSR #32`, `ASR #32`.
`ROR #0` means `RRX`

emu/src/cpu/single_data_transfer.rs
```diff
@@ -113,131 +104,232 @@ impl Arm7tdmi {
 
 #[cfg(test)]
 mod tests {
```
AlessioC31, line 106: Mhh, I don't think we should change all these tests like this. We're losing tests over the functionalities

AlessioC31 (changes requested):

emu/src/cpu/data_processing.rs
AlessioC31: Something like this:

```rust
 let amount = operand2.get_bits(7..=11);

                        if shift_type == ShiftKind::Ror && amount == 0 {
                            format!("R{rm}, RRX")
                        } else if shift_type == ShiftKind::Lsl && amount == 0 {
                            format!("R{rm}")
                        } else if amount == 0 {
                            format!("R{rm}, {shift_type} #32")
                        } else {
                            format!("R{rm}, {shift_type} #{amount}")
                        }
```

emu/src/cpu/instruction.rs
guerinoni: How can we calculate offset in ASM?

emu/src/cpu/instruction.rs
guerinoni: Sure, we can pick something like `SecondOperandKind` or `RegisterOrImmediateOperand`

emu/src/cpu/data_processing.rs
guerinoni: Right

emu/src/cpu/arm7tdmi.rs
guerinoni: Nope but is an additional test that before didn't exist :)

emu/src/cpu/instruction.rs
AlessioC31: I like the second one more, what do you think?

emu/src/cpu/instruction.rs
AlessioC31: If the shift is offset is immediate specified the offset is written in the opcode (`0..=11`), otherwise it is an `Rm` shifted by something and the shift part is similar to the one in data processing but without the register-specified amount

AlessioC31: But we can't calculate the offset in the `decode` because in the case of a Register specified, we should read the register and we don't do in the in the decode phase

emu/src/cpu/arm7tdmi.rs
AlessioC31: Sure but maybe we could add also the check for the pc and the status (arm/thumb) after the execution, if we don't have this test, once we're already touching it

emu/src/cpu/instruction.rs
```diff
@@ -186,7 +186,7 @@ impl From<u32> for ArmModeInstruction {
             let offset = op_code.get_bits(0..=23) << 2;
             Branch(condition, is_link, offset)
         } else if op_code.get_bits(26..=27) == 0b01 {
-            let op_kind: OperandKind = op_code.get_bit(25).into();
+            let op_kind: OperandKind = (!op_code.get_bit(25)).into();
```
AlessioC31, line 189: What do you think about adding a tiny comment here?

PR #156, opened 2022-12-27
Fix thumb ops

AlessioC31:
* We were not reading the value in `Rs` in ALU operations in THUMB mode
* We were not sign-extending the offset in the `BL` operation

Since we're using the sign-exteding feature in many places maybe we should put it somewhere? I've tried implementing in `Bits` but it's a bit of a mess since we would need `TryFrom`/`Into` `128` or something like that and it was not working, maybe I can try again, I tried couple of weeks ago


guerinoni (approved):
LGTM

PR #160, opened 2022-12-28
Progress on executing full BIOS

guerinoni:
Thumb: Partial implementation of `LoadStoreImmOffset`
CPU: Move into generic shift function `ASR`
Thumb: Implement `uncond_branch`
Thumb: Implement `orr`

### DIFF
--- a/emu/src/cpu/arm7tdmi.rs
+++ b/emu/src/cpu/arm7tdmi.rs
@@ -932,7 +932,12 @@ impl Arm7tdmi {
         let rd = op_code.get_bits(0..=2);
 
         match op {
-            ThumbModeAluInstruction::And => todo!(),
+            ThumbModeAluInstruction::And => self.and(
+                rd.into(),
+                self.registers.register_at(rd.try_into().unwrap()),
+                rs,
+                true,
+            ),
             ThumbModeAluInstruction::Eor => todo!(),
             ThumbModeAluInstruction::Lsl => todo!(),
             ThumbModeAluInstruction::Lsr => todo!(),
@@ -1784,6 +1789,23 @@ mod tests {
 
     #[test]
     fn check_alu_op() {
+        {
+            // and
+            let mut cpu = Arm7tdmi::default();
+            let op_code = 0b0100_0000_0001_1000;
+            let op_code: ThumbModeOpcode = cpu.decode(op_code);
+            assert_eq!(op_code.instruction, ThumbModeInstruction::AluOp);
+
+            cpu.cpsr.set_sign_flag(true);
+            cpu.cpsr.set_zero_flag(true);
+            cpu.registers.set_register_at(0, 1000);
+            cpu.registers.set_register_at(3, 8);
+            cpu.execute_thumb(op_code);
+
+            assert_eq!(cpu.registers.register_at(0), 8);
+            assert!(!cpu.cpsr.sign_flag());
+            assert!(!cpu.cpsr.zero_flag());
+        }
         {
             // tst
             let mut cpu = Arm7tdmi::default();
--- a/emu/src/cpu/arm7tdmi.rs
+++ b/emu/src/cpu/arm7tdmi.rs
@@ -127,7 +127,7 @@ impl Arm7tdmi {
             MultipleLoadStore => unimplemented!(),
             CondBranch => self.cond_branch(op_code),
             Swi => unimplemented!(),
-            UncondBranch => unimplemented!(),
+            UncondBranch => self.uncond_branch(op_code),
             LongBranchLink => self.long_branch_link(op_code),
         };
 
@@ -226,6 +226,16 @@ impl Arm7tdmi {
         } else {
             Some(SIZE_OF_THUMB_INSTRUCTION)
         }
+    }
+
+    fn uncond_branch(&mut self, op_code: ThumbModeOpcode) -> Option<u32> {
+        let offset = op_code.get_bits(0..=10) << 1;
+        let mask = 1 << 10;
+        let offset = (offset as i32 ^ mask) - mask;
+        let pc = self.registers.program_counter() as u32;
+        let new_pc = pc.wrapping_add(offset.try_into().unwrap());
+        self.registers.set_program_counter(new_pc);
+        Some(SIZE_OF_THUMB_INSTRUCTION)
     }
 
     fn add_subtract(&mut self, op_code: ThumbModeOpcode) -> Option<u32> {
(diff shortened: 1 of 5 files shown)


emu/src/cpu/arm7tdmi.rs
```diff
@@ -228,6 +228,14 @@ impl Arm7tdmi {
         }
     }
 
+    fn uncond_branch(&mut self, op_code: ThumbModeOpcode) -> Option<u32> {
+        let offset = (op_code.get_bits(0..=10) << 1) as i32;
```
AlessioC31, line 232: We need to sign-extend it. If we just do `as i32` it doesn't work.

Example in `u8/i8` where we want to sign-extend a 3-bits number:

```rust
let a: u8 = 0b110; // To sign-extend
let b: i8 = a as i8;
```

In this case `b == 0b0000_0110` while we would it to be `b == 0b1111_1110`.

To see how to sign-extend you can check the function `branch` for ARM. And while you're there maybe you can think how to move this logic somewhere so we don't copy paste it every time :P

emu/src/cpu/data_processing.rs
```diff
@@ -36,7 +36,7 @@ impl Arm7tdmi {
         // Shift Type (0=LSL, 1=LSR, 2=ASR, 3=ROR)
 
         let (result, carry) = match shift_type {
-            0 | 1 => {
+            0 | 1 | 2 => {
                 let a = alu_instruction::shift(
                     shift_amount.into(),
```
AlessioC31, line 41: Shouldn't it be `shift_type.into()` here?

emu/src/cpu/arm7tdmi.rs
```diff
@@ -308,6 +308,28 @@ impl Arm7tdmi {
...
 
+    fn load_store_immediate_offset(&mut self, op_code: ThumbModeOpcode) -> Option<u32> {
+        let byte_word: ReadWriteKind = op_code.get_bit(12).into();
+        let load_store: LoadStoreKind = op_code.get_bit(11).into();
+        let offset5 = op_code.get_bits(6..=10) as u32;
+        let rb = op_code.get_bits(3..=5);
+        let rd = op_code.get_bits(0..=2);
+
+        let base = self.registers.register_at(rb.try_into().unwrap());
+        let value = base.wrapping_add(offset5);
```
AlessioC31, line 319: I think `address` is better as name here to be consistent with other functions

emu/src/cpu/arm7tdmi.rs
```diff
@@ -308,6 +308,28 @@ impl Arm7tdmi {
...
+
+        match (load_store, byte_word) {
+            (LoadStoreKind::Store, ReadWriteKind::Word) => self
+                .registers
+                .set_register_at(rd.try_into().unwrap(), value),
```
AlessioC31, line 324: If it's a store we want to write into the memory, right?

AlessioC31 (changes requested):

emu/src/cpu/data_processing.rs
```diff
@@ -552,6 +552,13 @@ impl Arm7tdmi {
         self.cpsr.set_flags(add_result);
     }
 
+    pub fn thumb_orr(&mut self, rd: u32, rs: u32) {
+        let op = self.registers.register_at(rd.try_into().unwrap());
+        let result = op | rs;
+        self.registers
+            .set_register_at(rd.try_into().unwrap(), result);
+    }
+
```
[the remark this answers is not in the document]
guerinoni: Now I did, it's a little quirk in input parameter but 🤷🏻

emu/src/cpu/arm7tdmi.rs
guerinoni: Oh right right! Thanks!

emu/src/cpu/arm7tdmi.rs
guerinoni: Yes!

emu/src/cpu/data_processing.rs
guerinoni: Nothing change I guess :)

emu/src/cpu/arm7tdmi.rs
guerinoni: I also refactored :)

emu/src/cpu/data_processing.rs
AlessioC31: It does, no? `shift_type` is the type of the shift we want to perform. If we pass `shift_amount.into()` as a shift type we don't execute the correct shift

emu/src/cpu/arm7tdmi.rs
AlessioC31: You moved the logic somewhere else? I don't see it :(

emu/src/cpu/data_processing.rs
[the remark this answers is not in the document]
AlessioC31: What do you mean?

emu/src/cpu/data_processing.rs
guerinoni: Oh sorry yes! I miss read... This is the problem of doing to much TryFrom/From... Semantic error should be more explicit instead it is sneaky :(...
Fixed

emu/src/cpu/data_processing.rs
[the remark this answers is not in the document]
guerinoni: Because in arm uses 3 different input, in thumb just 2 but not important

emu/src/cpu/arm7tdmi.rs
guerinoni: Yes is a function `sign_extend`... Hit refresh :D

emu/src/cpu/arm7tdmi.rs
```diff
@@ -972,6 +999,12 @@ impl Arm7tdmi {
     }
 }
 
+/// Extends the bit at position `bit_to_extend` over a i32.
+const fn sign_extend(input: u32, bit_to_extend: u32) -> i32 {
+    let mask = 1 << bit_to_extend;
+    ((input as i32) ^ mask) - mask
+}
+
```
AlessioC31, line 1007: I think the `bit_to_extend` is a bit misleading. For example if we have a 23 bits two-complements number that we want to extend, we should pass `22` as parameter because we want the mask to be 23 bits long (thus shifting `1` by 22). Maybe we could call the parameter `number_of_bits` or something like that and use `number_of_bits - 1` to create the mask and call the functions with the correct number (`23` in line `974` for example).

AlessioC31 (changes requested):

PR #161, opened 2023-01-02
finish immediate transfer

guerinoni:


emu/src/cpu/arm7tdmi.rs
```diff
@@ -449,34 +449,46 @@ impl Arm7tdmi {
...
+                } else {
+                    self.registers
+                        .register_at(rd_source_destination_register as usize)
+                };
+
+                match transfer_type {
+                    HalfwordTransferType::UnsignedHalfwords => {
+                        if let Ok(mut mem) = self.memory.lock() {
+                            mem.write_at(address, value.get_bits(0..=7) as u8);
+                            mem.write_at(address + 1, value.get_bits(8..=15) as u8);
```
AlessioC31, line 466:
```suggestion
                            mem.write_half_word(address, value);
```

AlessioC31 (changes requested):

emu/src/cpu/arm7tdmi.rs
guerinoni: Nice, now we have that :)

AlessioC31 (approved):

PR #162, opened 2023-01-02
Thumb: Implement `mul` for ALU

guerinoni:

### DIFF
--- a/emu/src/cpu/data_processing.rs
+++ b/emu/src/cpu/data_processing.rs
@@ -578,6 +578,14 @@ impl Arm7tdmi {
             self.cpsr.set_zero_flag(result == 0);
         }
     }
+
+    pub fn mul(&mut self, reg_result: usize, op1: u32, op2: u32) {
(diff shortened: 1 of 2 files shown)


emu/src/cpu/data_processing.rs
```diff
@@ -578,6 +578,14 @@ impl Arm7tdmi {
             self.cpsr.set_zero_flag(result == 0);
         }
     }
+
+    pub fn mul(&mut self, reg_result: usize, op1: u32, op2: u32) {
+        let result = op1 * op2;
```
AlessioC31, line 583: This would fail in the case of overflow. We can execute the multiplication in `u64` and then take only the 32 lower bits

AlessioC31 (changes requested):

emu/src/cpu/data_processing.rs
guerinoni: Fixed!

AlessioC31 (approved):

PR #163, opened 2023-01-03
THUMB: Refactor and finish `HalfWord Load/Store`

AlessioC31:
* Refactored HalfWord LoadStore Register/Immediate offset into a single function.
* Implemented all cases


guerinoni (approved):
LGTM, very worth!

PR #164, opened 2023-01-03
UI: Add `breakpoints` window

AlessioC31:
Added breakpoints functionality to the UI


guerinoni:
Nice, but I would change something here...
1. Let's create an "handle" panel for start/stop/pause + breakpoints in order to keep separated from cpu register...
2. In CPU register we see value non in hex but adding a breakpoint it will took as hex
3. Not sure to make the check of register there but maybe we can call "pause" functionality for keeping more consistent? Or request pause when we are at that address??
4. Is not very clear that address is referring to PC, maybe we could add a breakpoint also for memory address... (next feature :P ) just make more clear

AlessioC31:
> Nice, but I would change something here...
> 
> 1. Let's create an "handle" panel for start/stop/pause + breakpoints in order to keep separated from cpu register...
> 2. In CPU register we see value non in hex but adding a breakpoint it will took as hex
> 3. Not sure to make the check of register there but maybe we can call "pause" functionality for keeping more consistent? Or request pause when we are at that address??
> 4. Is not very clear that address is referring to PC, maybe we could add a breakpoint also for memory address... (next feature :P ) just make more clear

1) Good idea
2) That's true I will solve that maybe showing register values in hex or giving the choice to the user
3) Isn't it what it's currently doing? When the address is reached I change the `bool` variable which is tecnically what the pause button does
4) I only knew about breakpoints in code execution. What are the breakpoints in the memory?

guerinoni:
Not sure if is a "standard" breakpoint but it can be useful stop when a certain memory address is read or write 🤷🏻

guerinoni (approved):

PR #165, opened 2023-01-09
Fix after debugging session

guerinoni:
- Arm: Fix start status of `cpsr`
- Thumb: Fix `uncond_branch`
- Interrupts: Add IE, IF, WAITCNT and one not used
- UI: Show register of CPU in hex


AlessioC31 (approved):

AlessioC31 (approved):

emu/src/render/ppu.rs
```diff
@@ -62,15 +62,14 @@ impl PixelProcessUnit {
             }
             3 => {
                 // Bitmap mode
-                for y in 0..LCD_HEIGHT {
-                    for x in 0..LCD_WIDTH {
-                        let index_color = (y * LCD_WIDTH + x) * 2;
+                for x in 0..LCD_HEIGHT {
+                    for y in 0..LCD_WIDTH {
+                        let index_color = (x * LCD_WIDTH + y) * 2;
```
AlessioC31, line 67: I remember we had a discussion about this and @gallottino said that in the "graphics world" `y` is used for height and `x` for width

emu/src/render/ppu.rs
```diff
@@ -81,22 +80,24 @@ impl PixelProcessUnit {
...
+                //         //     memory.video_ram[index_color],
+                //         //     memory.video_ram[index_color + 1],
+                //         // ]
+                //         // .into();
+                //         // FIXME
+                //         let color = Color(0);
+                //
+                //         gba_lcd.set_gbc_pixel(x, y, color);
+                //     }
+                // }
```
AlessioC31, line 100: Why commented?

emu/src/memory/internal_memory.rs
```diff
@@ -110,7 +110,7 @@ impl IoDevice for InternalMemory {
             0x04000200..=0x040003FE => self.interrupts.write_at(address, value),
             0x05000000..=0x050001FF => self.bg_palette_ram[address - 0x05000000] = value,
             0x05000200..=0x050003FF => self.obj_palette_ram[address - 0x05000200] = value,
-            0x06000000..=0x06017FFF => self.video_ram[address - 0x06000000] = value,
+            0x06000000..=0x06FFFFFF => self.video_ram[address - 0x06000000] = value,
```
AlessioC31, line 113: VRAM is ` 06000000-06017FFF   VRAM - Video RAM          (96 KBytes)` no?

emu/src/render/ppu.rs
guerinoni: Mmmm can be true, but is only temporary, this will be probably changed :)

emu/src/render/ppu.rs
guerinoni: Right now we are not using this mode, just for being sure of what we need time to time

emu/src/memory/internal_memory.rs
guerinoni: Yes, I extended to also NotUsed part because I want to avoid other problem of read/write on non used memory

AlessioC31: I think it would be better to put not_used explicitly instead of increasing the vram

AlessioC31 (approved):

AlessioC31 (approved):

PR #172, opened 2023-02-21
Memory: Add `Dma` struct

guerinoni:


AlessioC31 (approved):

emu/src/memory/dma.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    fn read_at(&self, address: Self::Address) -> Self::Value {
+        match address {
+            0x040000B0..=0x040000BB => self.banks[0].read_at(address - 0x040000B0),
+            0x040000BC..=0x040000C7 => self.banks[1].read_at(address - 0x040000BC),
+            0x040000C8..=0x040000D3 => self.banks[2].read_at(address - 0x040000C8),
+            0x040000D4..=0x040000DF => self.banks[3].read_at(address - 0x040000D4),
+            0x040000E0 => {
+                log("read in ununsed 0x040000E0");
+                0
```
AlessioC31, line 108: why hardcoded 0?

guerinoni: It shouldn't be used... Tried to put zero :)

AlessioC31: But in other memory regions we used hashmaps for these cases

guerinoni: I can but it is only 1... I hope is actually not used

AlessioC31: Maybe we can put a single variable, I don't know it was just to be "consistent", not mandatory I guess

PR #180, opened 2023-03-07
THUMB: Add `Load\Store sign-ext halfword`

AlessioC31:
Should we move `disassemble` function to a trait?

### DIFF
--- a/emu/src/cpu/arm7tdmi.rs
+++ b/emu/src/cpu/arm7tdmi.rs
@@ -149,6 +149,9 @@ impl Arm7tdmi {
     }
 
     pub fn execute_thumb(&mut self, op_code: ThumbModeOpcode) {
+        self.disassembler_buffer
+            .push(op_code.instruction.disassembler());
+
         use ThumbModeInstruction::*;
         let bytes_to_advance: Option<u32> = match op_code.instruction {
             MoveShiftedRegister => self.move_shifted_reg(op_code),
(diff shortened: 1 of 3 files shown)


emu/src/cpu/instruction.rs
```diff
@@ -288,7 +288,13 @@ pub enum ThumbModeInstruction {
     HiRegisterOpBX,
     PCRelativeLoad,
     LoadStoreRegisterOffset,
-    LoadStoreSignExtByteHalfword,
+    LoadStoreSignExtByteHalfword {
+        h: bool,
+        s: bool,
+        ro: u32,
+        rb: u32,
+        rd: u32,
```
guerinoni, line 296: Not very explicit name but 🤷🏻‍♂️

emu/src/cpu/arm7tdmi.rs
```diff
@@ -2141,4 +2196,100 @@ mod tests {
...
+                    rb: 1,
+                    rd: 2
+                }
+            );
+
+            cpu.registers.set_register_at(0, 10);
+            cpu.registers.set_register_at(1, 100);
+            cpu.registers.set_register_at(2, 0xF000_F0FF);
+
+            cpu.execute_thumb(op_code);
```
guerinoni, line 2221: We can split test for decoding from execution

guerinoni (commented):
Sure also we can create a trait for disassemble function :)

emu/src/cpu/instruction.rs
AlessioC31: Yep, changed thanks

emu/src/cpu/arm7tdmi.rs
AlessioC31: As discussed I did a first try with table tests, let's see

guerinoni (approved):

PR #183, opened 2023-03-09
Thumb: Implement load_address

guerinoni:

### DIFF
--- a/emu/src/cpu/arm7tdmi.rs
+++ b/emu/src/cpu/arm7tdmi.rs
@@ -177,7 +177,11 @@ impl Arm7tdmi {
             LoadStoreImmOffset => self.load_store_immediate_offset(op_code),
             LoadStoreHalfword => self.load_store_halfword(op_code),
             SPRelativeLoadStore => self.sp_relative_load_store(op_code),
-            LoadAddress => unimplemented!(),
+            LoadAddress {
+                sp,
+                r_destination,
+                offset,
+            } => self.load_address(sp, r_destination.try_into().unwrap(), offset),
             AddOffsetSP => self.add_offset_sp(op_code),
             PushPopReg => self.push_pop_register(op_code),
             MultipleLoadStore => unimplemented!(),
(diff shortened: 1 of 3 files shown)


emu/src/cpu/arm7tdmi.rs
```diff
@@ -1019,6 +1019,24 @@ impl Arm7tdmi {
...
 
+    fn load_address(&mut self, op_code: ThumbModeOpcode) -> Option<u32> {
+        let sp = op_code.get_bit(11);
+        let rd: usize = op_code.get_bits(8..=10).try_into().unwrap();
+        let word8: u32 = op_code.get_bits(0..=7).into();
+        let word8 = word8 << 2;
+
+        let v = if sp {
+            let stack_pointer = self.registers.register_at(REG_SP);
+            stack_pointer + word8
```
AlessioC31, line 1030: I think `wrapping_add` would be safer here (also 2 lines down)

emu/src/cpu/arm7tdmi.rs
```diff
@@ -1019,6 +1019,24 @@ impl Arm7tdmi {
...
+        } else {
+            let pc = self.registers.program_counter() as u32;
```
AlessioC31, line 1032: If we use `pc` we have to add `4`:

> Where the PC is used as the source register (SP = 0), bit 1 of the PC is always read as 0. The value of the PC will be 4 bytes greater than the address of the instruction before bit 1 is forced to 0.

We should also set the bit `1` to `0` since this is THUMB mode and it might happen that the bit 1 is `1` (since thumb instructions are 16 bits)

emu/src/cpu/arm7tdmi.rs
```diff
@@ -177,7 +177,7 @@ impl Arm7tdmi {
             LoadStoreImmOffset => self.load_store_immediate_offset(op_code),
             LoadStoreHalfword => self.load_store_halfword(op_code),
             SPRelativeLoadStore => self.sp_relative_load_store(op_code),
-            LoadAddress => unimplemented!(),
+            LoadAddress => self.load_address(op_code),
```
AlessioC31, line 180: Now that you introduced the separation between decode and execution I think that it would be better to write new instructions using that logic

guerinoni: I know, is a bit too effort now, I will refactor time by time :)

emu/src/cpu/arm7tdmi.rs
guerinoni: Done

emu/src/cpu/arm7tdmi.rs
guerinoni: Nice

PR #185, opened 2023-03-14
`PSR`: Wrong LE conditional check

rhighs:
Probably a typo here (`&&` is used instead of `||`) as the comment next to this statement also suggest using an `or`. 
Docs [here](https://rust-console.github.io/gbatek-gbaonly/#saved-program-status-registers-spsr_mode) just for reference.

Side note: we should implement some tests accounting for instructions with conditional execution, this way we can spot these little quirks more efficiently.


AlessioC31 (approved):
Well spotted! I wonder if this fixes the bug that we have right now

PR #186, opened 2023-05-13
CPU: Add program_counter to disassembler

RyanPaulGannon:
Added the Program Counter (`R15`) to the `disassembler` window to help with debugging.

There is an error where the Program Counter disappears after so many steps ...

### DIFF
--- a/emu/src/cpu/arm7tdmi.rs
+++ b/emu/src/cpu/arm7tdmi.rs
@@ -192,9 +192,13 @@ impl Arm7tdmi {
     }
 
     pub fn execute_thumb(&mut self, op_code: ThumbModeOpcode) {
-        self.disassembler_buffer
-            .push(op_code.instruction.disassembler());
-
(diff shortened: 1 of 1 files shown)


guerinoni:
I see the problem... Because we populate ASM buffer [here](https://github.com/RIP-Comm/clementine/blob/59e2902cf9fb02fd857af2827f2354f945e0a023/emu/src/cpu/arm7tdmi.rs#L94)
and [here](https://github.com/RIP-Comm/clementine/blob/59e2902cf9fb02fd857af2827f2354f945e0a023/emu/src/cpu/arm7tdmi.rs#L190)
But this is only the buffer that we print in UI.
I would suggest to check also [here](https://github.com/RIP-Comm/clementine/blob/59e2902cf9fb02fd857af2827f2354f945e0a023/ui/src/disassembler.rs#LL33C33-L33C33) if can be a better point or just add same logic for both mode. :)

guerinoni (approved):

PR #189, opened 2023-05-15
Thumb: Implement `MultipleLoadStore`

guerinoni:


emu/src/cpu/arm7tdmi.rs
```diff
@@ -1094,6 +1098,33 @@ impl Arm7tdmi {
...
+        base_register: u16,
+        register_list: u16,
+    ) -> Option<u32> {
+        match load_store {
+            LoadStoreKind::Store => unimplemented!("multiple store"),
+            LoadStoreKind::Load => {
+                let base_address = self.registers.register_at(base_register as usize);
+                for r in 0..=7 {
+                    if register_list.get_bit(r) {
+                        let address = base_address + (r as u32 * 4);
```
AlessioC31, line 1113: I'm not sure this is correct, imagine a register list like this: `0b00000010` and `R0 = 0x100`. We should then read at `0x100` and write in `R1`. But with this logic we read as `0x104`, right?

emu/src/cpu/instruction.rs
```diff
@@ -689,7 +693,25 @@ impl ThumbModeInstruction {
...
+            } => {
+                let instr = match load_store {
+                    LoadStoreKind::Load => "LDMIA",
+                    LoadStoreKind::Store => "STMIA",
+                };
+
+                let mut registers = String::new();
+                for i in 0..=7 {
+                    if register_list.get_bit(i) {
+                        registers.push_str(&format!("R{}, ", i));
```
AlessioC31, line 709: Like this we would have `{R0, R1, }`, right? (I mean at the end we would have an additional comma).

Also, to be consistent with the datasheet we should use `-` to separate registers :P

AlessioC31 (changes requested):

emu/src/cpu/instruction.rs
guerinoni: I know... Right now I took a shortcut :D we can leave as good first issue :) (another instruction has this problem) :D

emu/src/cpu/arm7tdmi.rs
```diff
@@ -1094,6 +1098,35 @@ impl Arm7tdmi {
...
+                        let address = base_address + (register_one as u32 * 4);
+                        let value = self.memory.lock().unwrap().read_word(address as usize);
+                        self.registers.set_register_at(r as usize, value);
+                        register_one += 1;
+                    }
+                }
+
+                let new_base_address = base_address + (register_list.count_ones() * 4);
+                self.registers
+                    .set_register_at(base_register as usize, new_base_address);
```
AlessioC31, line 1123: I know I'm nitpicking but I think it's more clear if we do something like:

```rust
let mut address = self.registers.register_at(...);

for ... {
  if ... {
     let value = ...;
     self.registers.set_registe_at(...);
     address += 4;
   }
  }

self.registers.set_register_at(...);
```

Or better, we could reuse the logic in the function `exec_data_transfer` in `arm7tdmi.rs`

AlessioC31 (changes requested):

emu/src/cpu/arm7tdmi.rs
guerinoni: Now is more easy to read :)

emu/src/cpu/arm7tdmi.rs
```diff
@@ -1094,6 +1098,33 @@ impl Arm7tdmi {
...
+                for r in 0..=7 {
+                    if register_list.get_bit(r) {
+                        let value = self.memory.lock().unwrap().read_word(address as usize);
+                        self.registers.set_register_at(r as usize, value);
+                        address += 4;
+                    }
+                }
+
+                let v = base_address + (register_list.count_ones() * 4);
+                self.registers.set_register_at(base_register, v);
```
AlessioC31, line 1121:
```suggestion
                self.registers.set_register_at(base_register, address);
```

Last one I swear

AlessioC31 (changes requested):

AlessioC31 (approved):

PR #192, opened 2023-05-18
Refactor code for CPU

guerinoni:
- Thumb: Implement load byte for `LoadStoreImmOffset`
- Thumb: Create dedicated module for thumb things
- Arm: Create dedicated module for arm things

Tests changes:
Under every mod we can add test for decoding or generating ASM, in this way in other tests we can simply check the execution and we can keep more tests but more clean and with a specific goal, instead every tests is too long to read.


emu/src/cpu/data_processing.rs
```diff
@@ -1,19 +1,15 @@
+use crate::bitwise::Bits;
```
AlessioC31, line 1: I think this file contains only the Dataprocessing instruction for Arm mode. Should we move this login into the new file `arm/alu_instructions.rs`?

emu/src/cpu/psr.rs
```diff
@@ -186,34 +186,6 @@ impl From<Psr> for u32 {
...
-
-impl From<bool> for CpuState {
-    fn from(state: bool) -> Self {
-        match state {
-            true => Self::Thumb,
-            false => Self::Arm,
-        }
-    }
-}
-
```
AlessioC31, line 216: I'm not sure about moving this to `cpu/cpu_modes.rs`. The CpuState is inside the PSR register, I've the feeling that it makes more sense to stay here but dunno

emu/src/cpu/data_processing.rs
AlessioC31: For me it's ok apart from this. But if you think we should tackle this in another PR I'm fine with it @guerinoni

PR #195, opened 2023-05-21
Implement more ALU operation

guerinoni:

### DIFF
--- a/emu/src/cpu/arm/operations.rs
+++ b/emu/src/cpu/arm/operations.rs
@@ -466,6 +466,11 @@ impl Arm7tdmi {
         self.cpsr.set_zero_flag(value == 0);
     }
 
+    /// Subtract the contents of rs from zero, and store the result in rd.
+    pub fn neg(&mut self, rd: usize, rs: u32) {
+        self.rsb(rd, rs, 0, true);
+    }
+
     fn teq(&mut self, rn: u32, op2: u32) {
         let value = rn ^ op2;
         self.cpsr.set_sign_flag(value.is_bit_on(31));
--- a/emu/src/cpu/arm/operations.rs
+++ b/emu/src/cpu/arm/operations.rs
@@ -6,7 +6,9 @@ use crate::cpu::arm::instructions::{SingleDataTransferKind, SingleDataTransferOf
 use crate::cpu::arm::mode::ArmModeOpcode;
 use crate::cpu::arm7tdmi::{Arm7tdmi, HalfwordTransferType};
 use crate::cpu::cpu_modes::Mode;
-use crate::cpu::flags::{Indexing, LoadStoreKind, Offsetting, OperandKind, ReadWriteKind};
+use crate::cpu::flags::{
+    Indexing, LoadStoreKind, Offsetting, OperandKind, ReadWriteKind, ShiftKind,
+};
 use crate::cpu::psr::CpuState;
 use crate::cpu::registers::REG_PROGRAM_COUNTER;
 use crate::memory::io_device::IoDevice;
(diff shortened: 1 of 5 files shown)


emu/src/cpu/arm/operations.rs
```diff
@@ -360,6 +360,20 @@ impl Arm7tdmi {
...
+        let rd_value = self.registers.register_at(rd);
+        let result = rd_value.rotate_right(rs);
+        let carry = rs.get_bit((rs - 1) as u8);
+        self.registers.set_register_at(rd, result);
+
+        self.cpsr.set_zero_flag(result == 0);
+        self.cpsr.set_sign_flag(result.get_bit(31));
+        self.cpsr.set_carry_flag(carry);
+    }
+
```
AlessioC31, line 376: I think we miss a bit of logic and we could maybe reuse the implementation that we already have for ROR.
In `shift_operand` in `emu/src/cpu/arm/operations.rs` we have something like:

```rust
match shift_type {
   // Shift Type (0=LSL, 1=LSR, 2=ASR, 3=ROR)
   0 | 1 | 2 => shift(...)
   3 => {
   // ROR logic
   }
```

Now... I don't know why we don't have the `ROR` in the `shift` function, if you look the `shift` function it has a `todo!` for `ROR`... Very weird.

By the way, I think we could take the opportunity to move the ROR shift logic (as you can see it has a couple of edge cases to it would be better to have this implementation only in one place) from `shift_operand` to `shift` and then we can reuse the `shift` function in the `ror` function implementation.

AlessioC31 (changes requested):

emu/src/cpu/arm/operations.rs
guerinoni: Right! I did a nice refactor :D

AlessioC31: Thanks!

Now we've another problem which I don't know how to solve :D

As you can see, in case `shift_amount == 0` in the `shift` function we have special cases like: `LSR#0 encodes LSR#32, ROR#0 encodes RRX`. Now... this is valid only if the shift amount is taken from immediate (because when we pass shift amount in the immediate we use the `#0` value as a special case to encode different things). In case the shift amount is passed via register, there's no reason to have this cases and if the byte read from the register is `0`, we simply don't apply shift and we don't change the carry flag.

Now... I see two options:

- Do a refactor of the entire shift logic between arm and thumb, like we could pass a new parameter to the `shift` function like `ShiftAmountSource` which encodes either `Register` or `Immediate` and if `Register` we don't do the special cases with `#0`. Right now this checks exists outside of the `shift` function: `get_operand` in `emu/src/cpu/arm/operations.rs` line `311`. But this `get_operand` is used only in the ARM part, in the thumb part we should have a similar `if` every time we call the `shift` function (and we take the shift amount from register, like this `ror`). So my idea was to move this if inside the `shift` so we have it only in one point.

- Add an `if` in the `fn ror` that you added that checks if the shift amount is 0, if it is the case it doesn't do any `shift`  and doesn't change the carry flag

guerinoni: For now I used a simple if in our `ror`... Let see if we can go forward and refactor step by step once we see all is working :D

emu/src/cpu/arm/operations.rs
```diff
@@ -360,6 +318,19 @@ impl Arm7tdmi {
...
 
+    /// Rotate Right rd by the value in rs, store the result in rd and set condition codes.
+    pub fn ror(&mut self, rd: usize, rs: u32) {
+        let rs = rs & 0xFF;
+        let rd_value = self.registers.register_at(rd);
+        if rs != 0 {
+            let r = shift(ShiftKind::Ror, rs, rd_value, self.cpsr.carry_flag());
+            self.registers.set_register_at(rd, r.result);
+            self.cpsr.set_zero_flag(r.result == 0);
+            self.cpsr.set_sign_flag(r.result.get_bit(31));
```
AlessioC31, line 329: These two we should still set even if the shifted amount is 0

emu/src/cpu/arm/operations.rs
```diff
@@ -463,6 +463,18 @@ impl Arm7tdmi {
...
+    pub fn neg(&mut self, rd: usize, rs: u32) {
+        let u_result = 0u32.overflowing_sub(rs);
+        let i_result = 0i32.overflowing_sub(rs as i32);
+
+        self.registers.set_register_at(rd, u_result.0);
+        self.cpsr.set_carry_flag(!u_result.1);
+        self.cpsr.set_overflow_flag(i_result.1);
+        self.cpsr.set_zero_flag(u_result.0 == 0);
+        self.cpsr.set_sign_flag(i_result.0 > 0);
+    }
```
AlessioC31, line 476: I think here we can reuse `rsb(rd, rs, 0)`

AlessioC31 (changes requested):

emu/src/cpu/arm/operations.rs
guerinoni: Done

emu/src/cpu/arm/operations.rs
guerinoni: RSB does a SUB... I don't think is the semantically

AlessioC31: `neg` is the same as an `rsbs` in arm 

image

guerinoni: Fixed :D, at some point we should extract ALU

AlessioC31 (approved):

PR #196, opened 2023-05-22
Arm: Implement mul opcodes (halfwords and non-halfwords)

rhighs:
This PR brings the implementation of the various opcodes for multiplications in arm-mode, I decided to keep the already present `ArmModeInstruction::Multiply` type and extending it a little with a `ArmModeMultiplyKind` to allow for a basic way of differentiating the opcodes.

Overflows are checked using `u64|u32::overflowing_mul(op1, op2)` which conveniently returns a tuple containing the mul result and whether an overflow occurred. I chose this way as it clearly shows the intent of what's happening there, perhaps we could only use such function when the `S` flag is set, otherwise there's really no point in using that. (It might bring extra overhead for no reason, but you guys tell me if that's any good to keep it like so).

Other than that, I don't really have anything important here to address. The opcodes are quite similar to each other, and it's not guaranteed we're gonna need them all, but still I brought them as it required little effort to do so.

As for the testing side I tried to test the most common case for each one, some need to check for condition flags, some don't and so on...

Looking forward to improve this as much as we can! :v:

### DIFF
--- a/emu/src/cpu/arm/instructions.rs
+++ b/emu/src/cpu/arm/instructions.rs
@@ -81,13 +81,29 @@ pub enum ArmModeInstruction {
         destination: u32,
         op2: AluSecondOperandInfo,
     },
+    Multiply {
+        variant: ArmModeMultiplyVariant,
+        condition: Condition,
+        should_set_codes: bool,
+        rd_destination_register: u32,
+        rn_accumulate_register: u32,
+        rs_operand_register: u32,
+        rm_operand_register: u32,
+    },
+    MultiplyLong {
+        variant: ArmModeMultiplyLongVariant,
+        condition: Condition,
+        should_set_codes: bool,
+        rdhi_destination_register: u32,
+        rdlo_destination_register: u32,
+        rs_operand_register: u32,
+        rm_operand_register: u32,
+    },
     PSRTransfer {
         condition: Condition,
         psr_kind: PsrKind,
         kind: PsrOpKind,
     },
-    Multiply,
-    MultiplyLong,
     SingleDataSwap,
     BranchAndExchange {
         condition: Condition,
@@ -148,6 +164,56 @@ pub enum ArmModeInstruction {
     SoftwareInterrupt,
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum ArmModeMultiplyVariant {
+    Mul,
+    Mla,
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum ArmModeMultiplyLongVariant {
+    Umull,
+    Umlal,
+    Smull,
+    Smlal,
+}
+
+impl From<u32> for ArmModeMultiplyVariant {
+    fn from(op_code: u32) -> Self {
+        let mul_op_code: u32 = op_code.get_bits(21..=24);
+        match mul_op_code {
+            0b0000 => Self::Mul,
+            0b0001 => Self::Mla,
+            _ => unreachable!(),
+        }
+    }
+}
+
+impl std::fmt::Display for ArmModeMultiplyVariant {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
+impl From<u32> for ArmModeMultiplyLongVariant {
+    fn from(op_code: u32) -> Self {
+        let mul_op_code: u32 = op_code.get_bits(21..=24);
+        match mul_op_code {
+            0b0100 => Self::Umull,
+            0b0101 => Self::Umlal,
+            0b0110 => Self::Smull,
+            0b0111 => Self::Smlal,
+            _ => unreachable!(),
+        }
+    }
+}
+
+impl std::fmt::Display for ArmModeMultiplyLongVariant {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
 impl ArmModeInstruction {
     pub(crate) fn disassembler(&self) -> String {
         match self {
@@ -186,7 +252,34 @@ impl ArmModeInstruction {
                         format!("{alu_instruction}{condition}{set_string} R{destination}, {op2}")
                     }
                 }
-            }
+            },
+            Self::Multiply {
+                variant,
+                condition,
+                should_set_codes,
+                rd_destination_register,
+                rn_accumulate_register,
+                rs_operand_register,
+                rm_operand_register,
+            } => {
+                match variant {
+                    ArmModeMultiplyVariant::Mul =>
+                        format!("MUL{condition}{should_set_codes} {rd_destination_register}, {rm_operand_register}, {rs_operand_register}"),
+                    ArmModeMultiplyVariant::Mla =>
+                        format!("MLA{condition}{should_set_codes} {rd_destination_register}, {rm_operand_register}, {rs_operand_register}, {rn_accumulate_register}"),
+                }
+            },
+            Self::MultiplyLong {
+                variant,
+                condition,
+                should_set_codes,
+                rdhi_destination_register,
+                rdlo_destination_register,
+                rs_operand_register,
+                rm_operand_register,
+            } => {
+                format!("{variant}{condition}{should_set_codes} {rdlo_destination_register}, {rdhi_destination_register}, {rm_operand_register}, {rs_operand_register}")
+            },
             Self::PSRTransfer {
                 condition,
                 psr_kind,
@@ -204,8 +297,6 @@ impl ArmModeInstruction {
                     format!("MSR{condition} {psr_kind}_flg, {operand}")
                 }
             },
-            Self::Multiply => "".to_owned(),
-            Self::MultiplyLong => "".to_owned(),
             Self::SingleDataSwap => "".to_owned(),
             Self::BranchAndExchange {
                 condition,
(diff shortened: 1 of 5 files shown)


emu/src/bitwise.rs
```diff
@@ -220,12 +264,37 @@ mod tests {
 
     #[test]
     fn set_bit() {
-        let mut b = 0b1100110_u32;
+        let mut b = 0b10001001_u32;
         b.set_bit(0, true);
         b.set_bit(1, true);
         b.set_bit(2, false);
         b.set_bit(3, false);
-        assert_eq!(b, 0b1100011)
```
AlessioC31, line 228: I think we can revert this change right?

emu/src/cpu/arm/instructions.rs
```diff
@@ -77,8 +77,7 @@ pub enum ArmModeInstruction {
         destination: u32,
         op2: AluSecondOperandInfo,
     },
-    Multiply,
-    MultiplyLong,
+    Multiply(ArmModeMultiplyKind),
```
AlessioC31, line 80: I'm not 100% sure about this, just because semantically the `ArmModeInstruction` maps every possible opcode as per armv7tdmi datasheet.

Do you think we could keep the old `Multiply` and `MultiplyLong` or are there any specific reason (maybe it's easier to implement like this?) to keep things under `Multiply` only?


cc @guerinoni

AlessioC31 (commented):
Thanks a lot! This looks like a ton of work :D

This is a first fast review, I will look deeper into implementation later

emu/src/bitwise.rs
rhighs: Yeah I kind of agree on this, I really only use `set_bits` testing for testing purposes I find it handy, but i can change this quite easily

AlessioC31: Sorry what I meant is: why are we changing the test of `set_bit`? We're not touching it in this PR no?

rhighs: Oh ok, I have no idea how that changed :joy: , I'll revert it np

emu/src/cpu/arm/instructions.rs
rhighs: Also, I put tests for muls in `emu/src/cpu/arm7tdmi.rs` perhaps I should move them to `emu/src/cpu/arm/operations.rs`, right?

emu/src/cpu/arm/instructions.rs
```diff
@@ -77,8 +77,8 @@ pub enum ArmModeInstruction {
         destination: u32,
         op2: AluSecondOperandInfo,
     },
-    Multiply,
-    MultiplyLong,
+    Multiply(ArmModeMultiplyVariant),
+    MultiplyLong(ArmModeMultiplyLongVariant),
```
guerinoni, line 81: I would change this as a struct variant like other

emu/src/bitwise.rs
```diff
@@ -55,6 +55,50 @@ where
...
+        //     0b0000....00000100_10000000_u128 =
+        //     --------------------------------
+        //     0b0000....10010100_11110011_u128
+
+        <Self as TryFrom<u128>>::try_from((self_bits & reverse_mask) | value_bits).unwrap()
+    }
+
+    fn set_bits_mut(&mut self, bits_range: RangeInclusive<u8>, value: Self) {
+        self.clone_from(&self.set_bits(bits_range, value))
+    }
```
guerinoni, line 100: This is the real `set_bit` because it changes the value in the same instance where you call method. The other returns a copy with some bits changed...

AlessioC31: I would try to be consistent with the existing methods. For example we only have `set_bit`, `set_byte`, etc... which change the instance on which are called, without returning `Self`.

I mean we can have this behaviour but I would introduce it in every function, just to be consistent (would it make the usage better?)

rhighs: Oh ok, I get it now, it makes sense now I guess. Gonna change it np :+1:

emu/src/cpu/arm/operations.rs
```diff
@@ -841,6 +836,293 @@ impl Arm7tdmi {
...
+    pub fn multiply(&mut self, mul_variant: ArmModeMultiplyVariant, op_code: ArmModeOpcode) {
+        use ArmModeMultiplyVariant::*;
+
+        match mul_variant {
+            // Unsiged multiply (32-bit by 32-bit, bottom 32-bit result).
+            Mul => self.mul(op_code),
+            // Unsiged multiply-accumulate (32-bit by 32-bit, bottom 32-bit accumulate and result).
+            Mla => self.mla(op_code),
+        }
+    }
```
guerinoni, line 878: I think we removed this right? @AlessioC31

AlessioC31: Yep, [GBA doesn't have a coprocessor](https://github.com/RIP-Comm/clementine/commit/7c895e703dd1611e65bf46acd20a7ab33e8a2f88)

rhighs: Looking good now? @guerinoni @AlessioC31

guerinoni (approved):
LGTM

emu/src/cpu/arm/operations.rs
```diff
@@ -841,6 +836,264 @@ impl Arm7tdmi {
...
+            // Signed long multiply (32-bit by 32-bit, 64-bit result)
+            Smull => self.smull(op_code),
+            // Signed multiply-accumulate (32-bit by 32-bit, 64-bit accumulate and result)
+            Smlal => self.smlal(op_code),
+        }
+    }
+
+    pub fn mul(&mut self, op_code: ArmModeOpcode) {
+        let set_condition_codes = op_code.get_bit(20);
+        assert!(0b1001u32 == op_code.get_bits(4..=7));
```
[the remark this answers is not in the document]
AlessioC31: We could instead assert that we do not use R15 as destination / operand and that rd != rm but you choose :D

emu/src/cpu/arm/operations.rs
```diff
@@ -841,6 +836,264 @@ impl Arm7tdmi {
...
+        let rdlo_destination_register = op_code.get_bits(12..=15);
+
+        let rm_operand_value = self
+            .registers
+            .register_at(rm_operand_register.try_into().unwrap());
+        let rs_operand_value = self
+            .registers
+            .register_at(rs_operand_register.try_into().unwrap());
+
+        let rm_operand_value_sgn: i64 = i32::from_be_bytes(rm_operand_value.to_be_bytes()) as i64;
```
AlessioC31, line 1025: Doesn't something like `rm_operand_value as i64` work?

AlessioC31: (same in similar cases)

AlessioC31 (changes requested):
I'm sorry for the time taken to review this. Overall it looks fine, there's something I think we should change but nothing too big, thanks!

Some comments are related to the fact that it would be nice to try to stick to a single "code style". We're not that many contributing but I think we should aim at this in any case

emu/src/cpu/arm/operations.rs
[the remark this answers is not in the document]
rhighs: I like assertions, they're an easy way of communicating a requirement to the reader, I'll go for the second one :P

emu/src/cpu/arm/operations.rs
```diff
@@ -841,6 +836,264 @@ impl Arm7tdmi {
...
+        self.registers
+            .set_register_at(rd_destination_register.try_into().unwrap(), result.0);
+
+        if set_condition_codes {
+            let any_overflow = result_mul.1 || result.1;
+            self.cpsr.set_carry_flag(any_overflow);
+            self.cpsr.set_overflow_flag(any_overflow);
+            self.cpsr.set_zero_flag(result.0 == 0);
+        }
+    }
```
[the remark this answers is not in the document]
rhighs: Sure I'll into that :+1:

emu/src/cpu/arm/operations.rs
```diff
@@ -841,6 +836,264 @@ impl Arm7tdmi {
...
+            .registers
+            .register_at(rs_operand_register.try_into().unwrap());
+
+        let result = u32::overflowing_mul(rm_operand_value, rs_operand_value);
+        self.registers
+            .set_register_at(rd_destination_register.try_into().unwrap(), result.0);
+
+        if set_condition_codes {
+            self.cpsr.set_carry_flag(result.1);
+            self.cpsr.set_overflow_flag(result.1);
```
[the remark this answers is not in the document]
rhighs: I must have missed this part, thanks :+1:

emu/src/cpu/arm/operations.rs
AlessioC31: Thinking about this a bit more, would this line work?

I mean, will it do the sign-extension on the bits?

Imagine the output of `from_be_bytes` is a negative number (so it has a 1 as bit 31), when we do `as i64`, would this sign be extended (putting `1` to all the new 32 bits to the left)?

(There's the same problem with the `rm_operand_value as i64` I suggested before)

We have `sign_extend` for unsigned types if needed

AlessioC31: image
image

Ok it works, nice :D

rhighs: I'm not quite sure that would work the same, as I need to keep the sign going from 32 to 64 bits, perhaps it might be the case to use the sign extension utility already available under `bitwise.rs`.
 
Simply doing `(rs_operand_value as i32) as i64` does the trick, but you guys tell me :)

AlessioC31: > I'm not quite sure that would work the same, as I need to preserve the sign value going from 32 to 64 bits, perhaps it might be the case to use the sign extension utility already available under `bitwise.rs`.
> 
> Simply doing `(rs_operand_value as i32) as i64` does the trick, but you guys tell me :)

I think what you suggested (`(rs_operand_value as i32) as i64`) is what I like most but given that we can't really avoid the double cast you can choose whatever you like most

rhighs: In fact `to_be -> from_be` is a bit silly, I probably went down this way because of the implementations of other opcodes that we've removed

emu/src/cpu/arm/operations.rs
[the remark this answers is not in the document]
rhighs: I don't quite understand what is meant by `The C (Carry) flag is set to a meaningless`, what is it practically?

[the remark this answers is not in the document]
AlessioC31: What I understood is that the instruction does not guarantee to leave the `C` flag untouched, so the following instructions cannot use this value since it would be "undefined" in some way.

I guess that for the sake of the emulator we can leave it untouched

PR #197, opened 2023-05-22
UI: Use a fixed vector to keep ASM generated

guerinoni:
In this way UI is very fluid, before was very laggy at some point,
because of the string that need to be render for all ASM generated since
time 0. Now I created a little data struct that answer to our need in
order to keep only the last 1k instructions. Seems a reasonable number.
For future improvement we can enable also a possibility to save as file.

### DIFF
--- a/vecfixed/src/lib.rs
+++ b/vecfixed/src/lib.rs
@@ -1,0 +1,112 @@
+use std::collections::VecDeque;
+
+/// VecFixed is basically a vector that keep a fixed size. Every time new element is pushed
+/// to the vector, the oldest element is removed and the latest pushed is added to the end.
+#[derive(Default)]
+pub struct VecFixed<const N: usize, T: Default + ToString> {
+    next_index: usize,
+    buffer: VecDeque<T>,
+}
+
+impl<const N: usize, T: Default + ToString> VecFixed<N, T> {
+    pub fn new() -> Self {
+        // let mut buffer = ;
+        // for _ in 0..N {
+        //     buffer.push_back(T::default());
+        // }
+
+        Self {
+            next_index: 0,
+            buffer: VecDeque::with_capacity(N),
+        }
+    }
+
+    pub fn push(&mut self, element: T) {
+        if self.next_index == N {
+            self.buffer.pop_front();
+        } else {
+            self.next_index += 1;
+        }
+
+        self.buffer.push_back(element);
+    }
+
+    /// Join the elements of the VecFixed buffer into a string.
+    pub fn join(&self, separator: &str) -> String {
+        if self.buffer.is_empty() {
+            return String::new();
+        }
+
+        let mut s = String::new();
+        s.push_str(&self.buffer[0].to_string());
+
+        for element in self.buffer.iter().skip(1) {
+            s.push_str(separator);
+            s.push_str(&element.to_string());
+        }
+
+        s
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn new() {
+        let ring: VecFixed<10, u8> = VecFixed::new();
+        assert_eq!(ring.next_index, 0);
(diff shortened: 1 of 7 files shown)


vecfixed/src/lib.rs
```diff
@@ -0,0 +1,107 @@
+//noinspection ALL
+/// VecFixed is basically a vector that keep a fixed size. Every time new element is pushed
+/// to the vector, the oldest element is removed and the latest pushed is added to the end.
+#[derive(Default)]
+pub struct VecFixed<const N: usize, T: Default + ToString> {
+    start: usize,
```
AlessioC31, line 6: What do you think about changing `start` to something like `next_index`?

vecfixed/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+    buffer: Vec<T>,
+}
+
+impl<const N: usize, T: Default + ToString> VecFixed<N, T> {
+    pub fn new() -> Self {
+        let mut buffer = Vec::with_capacity(N);
+        for _ in 0..N {
+            buffer.push(T::default());
+        }
```
AlessioC31, line 15: Curiosity, doesn't `let mut buffer = vec![T::default(); N]` work?

vecfixed/src/lib.rs
```diff
@@ -0,0 +1,107 @@
+//noinspection ALL
+/// VecFixed is basically a vector that keep a fixed size. Every time new element is pushed
+/// to the vector, the oldest element is removed and the latest pushed is added to the end.
+#[derive(Default)]
+pub struct VecFixed<const N: usize, T: Default + ToString> {
+    start: usize,
+    buffer: Vec<T>,
```
AlessioC31, line 7: What about using a `VecDeque` here? Since we want to drop from the left and push to the right, it would be better in terms of performances. `rotate_left` for `Vec` is O(n). While it's O(1) for `VecDeque` (in cases like ours where we rotate by 1). Or if we use `VecDeque` we also have `pop_front` if we don't want to rotate, so we could do 
```rust
pop_front();
push_back(elem);
```

AlessioC31 (changes requested):

vecfixed/src/lib.rs
guerinoni: Nice

vecfixed/src/lib.rs
guerinoni: We can also do it, but this requires `Clone` to be satisfied

AlessioC31: Uh ok makes sense

vecfixed/src/lib.rs
AlessioC31: Does it? We can instantiate a `VecDeque::with_capacity` and then once we are at length == N we pop from the back and we push to the front

AlessioC31: Ok I've checked the source code:

image

So it only grows when it's full, in our case it should not be the case

guerinoni: Yes but you still need to handle manually the shift and deletions of elements... More or less is same as this :)

AlessioC31: image

I think in our case it should be O(1) since `mid` for us is constant.

It will probably not change much in terms of performances (or will it?) but since we're writing our own boundedqueue, we can try to optimize it as much as reasonable

guerinoni: Yes seems a good improvement, I can change it :)

guerinoni:
@AlessioC31 
The new implementation benchmark
```
vector_fixed            time:   [8.0615 µs 8.0822 µs 8.1040 µs]
                        change: [-57.499% -57.326% -57.162%] (p = 0.00 < 0.05)
                        Performance has improved.
```

AlessioC31 (approved):

PR #198, opened 2023-05-23
Pipelining

AlessioC31:
Well, it's time.

Only a problem: it doesn't work anymore (it executes fine until a certain point, then something happens)

Commit message because I want to go to bed, I will split it tomorrow :D


AlessioC31:
Fixed, it was a missed `flush` :D

guerinoni (approved):
Super

PR #199, opened 2023-05-30
ARM: Extract `PSRTransfer` instruction from `DataProcessing`

AlessioC31:
These two type of instructions do very different
things even if they're under the same opcode
in the datasheet.
Splitting them makes also easier to
manage disassembly for PSR.

### DIFF
--- a/emu/src/cpu/arm/alu_instruction.rs
+++ b/emu/src/cpu/arm/alu_instruction.rs
@@ -204,13 +204,14 @@ pub fn shift(kind: ShiftKind, shift_amount: u32, rm: u32, carry: bool) -> Arithm
 }
 
 /// Represents the kind of PSR operation
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub enum PsrOpKind {
     /// MSR operation (transfer PSR contents to a register)
-    Mrs,
+    Mrs { destination_register: u32 },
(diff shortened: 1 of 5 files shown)


emu/src/cpu/arm/alu_instruction.rs
```diff
@@ -219,23 +220,66 @@ impl From<u32> for PsrOpKind {
...
         } else if op_code.get_bits(26..=27) == 0b00
             && op_code.get_bits(23..=24) == 0b10
             && op_code.get_bits(12..=21) == 0b10_1000_1111
         {
-            Self::MsrFlg
+            Self::MsrFlg {
+                operand: if op_code.get_bit(25) {
+                    AluSecondOperandInfo::Immediate {
+                        base: op_code.get_bits(0..=7),
+                        shift: op_code.get_bits(8..=11) * 2,
```
guerinoni, line 241: `* 2`? We usually use << N

guerinoni (approved):
It's ok, just a little doubt

emu/src/cpu/arm/alu_instruction.rs
AlessioC31: I used the same logic as in `emu/src/cpu/arm/instructions.rs` line `472`

guerinoni (approved):
LGTM

PR #200, opened 2023-05-31
Halfwordtransfer asm

AlessioC31:
Left to do:
- Add tests for the disassembler string


emu/src/cpu/arm7tdmi.rs
```diff
@@ -466,13 +466,13 @@ impl Arm7tdmi {
     }
 }
 
-pub enum HalfwordTransferType {
+pub enum HalfwordTransferKind {
```
guerinoni, line 484: This is good

guerinoni (approved):

PR #205, opened 2023-07-19
Initial work for interrupts

AlessioC31:


emu/src/bus.rs
```diff
@@ -32,12 +32,18 @@ impl Bus {
...
         self.cycles_count += 1;
 
         // TODO: move this somewhere in the UI
         log(format!("CPU Cycles: {}", self.cycles_count));
 
         // Step ppu, dma, interrupts, timers, etc...
+        let mut mem = self.internal_memory.lock().unwrap();
+
+        let val = *mem.interrupts.interrupt_request.back().unwrap();
+        mem.interrupts.interrupt_request.push(val);
```
guerinoni, line 46: I don't fully understand this, removing from back and pushing to front... 🤔

emu/src/memory/interrupts.rs
```diff
@@ -49,7 +55,7 @@ impl Interrupts {
 
         Self {
             interrupt_enable: IORegister::with_access_control(ReadWrite),
-            interrupt_request: IORegister::with_access_control(ReadWrite),
+            interrupt_request: VecFixed::initialize(0),
```
guerinoni, line 58: Why don't keep `new` for this?

emu/src/bus.rs
AlessioC31: With the first line I don't want to remove from the back, the behaviour I'm trying to have is:

`interrupt_request = [1, 0, 0, 0, 3]`

I execute these two lines and I want:

`interrupt_request = [1, 1, 0, 0, 0]`

AlessioC31: So I want to take the back and push a copy of it

AlessioC31: The reason is that once I write to the interrupt requests register, it is delayed by 4 cycles before the CPU sees it.

So my idea was to write the back when we write the memory and the CPU sees the front of this buffer.
Every step we propagate the information so:

`T = 0`: `interrupt_request = [0, 0, 0, 0, 1]` (requesting an interrupt like hblank, vblank or something like that)
`T = 1`: `interrupt_request = [0, 0, 0, 1, 1]`
`T = 2`: `interrupt_request = [0, 0, 1, 1, 1]` 
`T = 3`: `interrupt_request = [0, 1, 1, 1, 1]`
`T = 4`: `interrupt_request = [1, 1, 1, 1, 1]`  

So the CPU sees the `1` at `T=4`

emu/src/memory/interrupts.rs
AlessioC31: I wanted to initialize every element of the ring at `0`. With `new` I would have an empty ring with some capacity

emu/src/bus.rs
guerinoni: Mmm ok, got it thanks...

emu/src/memory/interrupts.rs
guerinoni: Make sense

guerinoni (approved):

PR #207, opened 2023-07-28
Avoid having `Arc<Mutex<Bus>>` inside `Arm7tdmi`

AlessioC31:
The main idea of this change is to avoid having `Arc<Mutex<Bus>>` inside `Arm7tdmi` and have just a `Bus` instead.
Having an ArcMutex it was causing me some issues while working on interrupts implementation. Plus it's always been something I was not 100% sure about. I changed the `Gba` logic to hold only a reference to the cpu instead of both cpu and bus.

### DIFF
--- a/emu/src/cpu/arm/operations.rs
+++ b/emu/src/cpu/arm/operations.rs
@@ -15,7 +15,6 @@ use crate::cpu::flags::{
 };
 use crate::cpu::psr::CpuState;
 use crate::cpu::registers::REG_PROGRAM_COUNTER;
-use crate::memory::io_device::IoDevice;
 use logger::log;
 
 use super::alu_instruction::{AluSecondOperandInfo, PsrKind};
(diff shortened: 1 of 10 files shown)


emu/src/cpu/arm/operations.rs
```diff
@@ -556,7 +555,7 @@ impl Arm7tdmi {
             Indexing::Post => address.try_into().unwrap(),
         };
 
-        let mut bus = self.bus.lock().unwrap();
+        // let mut bus = self.bus.lock().unwrap();
```
guerinoni, line 558: ?

ui/src/cpu_registers.rs
```diff
@@ -48,10 +48,19 @@ impl UiTool for CpuRegisters {
...
         // If it's poisoned it means that the thread that executes instructions
         // panicked, we still want access to registers to debug
-        let registers = self.gba.lock().map_or_else(
-            |poisoned| poisoned.into_inner().cpu.registers.to_vec(),
-            |gba| gba.cpu.registers.to_vec(),
-        );
+        // let registers = self.gba.lock().map_or_else(
+        //     |poisoned| poisoned.into_inner().cpu.lock().unwrap().registers.to_vec(),
+        //     |gba| gba.cpu.lockregisters.to_vec(),
+        // );
```
guerinoni, line 54: ?

guerinoni (approved):
Maybe some leftover comments

PR #208, opened 2023-07-31
Handle exceptions

AlessioC31:
First implementation of exceptions.

Draft since nothing fires an interrupt yet so I don't know if it works :D

### DIFF
--- a/emu/src/cpu/arm7tdmi.rs
+++ b/emu/src/cpu/arm7tdmi.rs
@@ -34,6 +34,76 @@ pub struct Arm7tdmi {
     decoded_thumb: Option<ThumbModeOpcode>,
 }
 
+#[derive(Copy, Clone)]
+#[allow(dead_code)]
+enum ExceptionType {
+    Reset,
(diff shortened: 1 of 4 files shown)


emu/src/cpu/arm7tdmi.rs
```diff
@@ -34,6 +34,76 @@ pub struct Arm7tdmi {
...
+impl ExceptionType {
+    pub const fn address(&self) -> usize {
+        match *self {
+            Self::Reset => 0x0,
+            Self::UndefinedInstruction => 0x4,
+            Self::SoftwareInterrupt => 0x8,
+            Self::PrefetchAbort => 0xC,
+            Self::DataAbort => 0x10,
+            Self::Irq => 0x18,
+            Self::Fiq => 0x1C,
```
guerinoni, line 58: Where those values come from?

AlessioC31: I took it from the ARM datasheet:

image

PR #210, opened 2023-08-03
Move `Lcd` out of `InternalMemory`

AlessioC31:
The idea of this PR is to propose a new way to manage memory and hardware components.
Before we had everything under `InternalMemory` but every field was just containing registers.
This is not suitable now that we approach implementing the LCD since it has also its login instead of just a bunch of registers.

Also, having everything under `InternalMemory` was not "semantically" correct in my opinion, hardware components are external components to the CPU.

If we agree with this type of refactoring, we would also get rid of the `IORegister` struct which is useless at this point. I added it many months ago to try to handle read/write access through it but then we saw that it was not possible.

To manage read/write access we could do something like this:

- If the register is R/W, just make the field `pub`
- If the register is R, keep the field private and write a getter for the field
- If the register is W, just make the field `pub`

I also changed a bit the read/write functions in `Bus`.
Now we have `write_raw` and `read_raw` that map every address to the correct component and we have wrappers like `read_word`, `write_word`, `read_half_word`, etc... that should calculate how many wait cycles the CPU has to be stopped (depending on wait cycles and bus width of the component).

### DIFF
--- a/emu/src/memory/internal_memory.rs
+++ b/emu/src/memory/internal_memory.rs
@@ -5,7 +5,6 @@ use logger::log;
 use crate::bitwise::Bits;
 use crate::memory::dma::Dma;
 use crate::memory::io_device::IoDevice;
-use crate::memory::lcd_registers::LCDRegisters;
 use crate::memory::timer_registers::TimerRegisters;
 
 use super::interrupts::Interrupts;
@@ -22,9 +21,6 @@ pub struct InternalMemory {
 
     /// From 0x03000000 to 0x03007FFF (32kb).
     working_iram: Vec<u8>,
-
-    /// From 0x04000000 to 0x04000055 (0x56 bytes).
-    pub lcd_registers: LCDRegisters,
 
     /// From 0x04000060 to 0x040000AF
     sound: Sound,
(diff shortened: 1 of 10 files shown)


emu/src/memory/internal_memory.rs
```diff
@@ -151,7 +132,7 @@ impl IoDevice for InternalMemory {
             0x00000000..=0x00003FFF => self.bios_system_rom[address],
             0x02000000..=0x0203FFFF => self.working_ram[address - 0x02000000],
             0x03000000..=0x03007FFF => self.working_iram[address - 0x03000000],
-            0x04000000..=0x04000055 => self.lcd_registers.read_at(address),
+            0x04000000..=0x04000055 => unreachable!(),
```
guerinoni, line 135: I think we can remove this... Since we should not go there

guerinoni (approved):
LGTM, left a couple of comments... Also we should consider the trait 
`pub trait IoDevice {` Is this still useful? Or should we change it?

emu/src/memory/internal_memory.rs
AlessioC31: You're right, thanks

AlessioC31:
> LGTM, left a couple of comments... Also we should consider the trait `pub trait IoDevice {` Is this still useful? Or should we change it?

Thanks, I think that once we move to this new approach the `IoDevice` trait will be unused

PR #214, opened 2023-09-21
Create `logger` and `disassembler` feature default disabled

guerinoni:
- Just: Remove useless cmds
- Just: Add quick cmd for run rom
- Logger: Make logger a feature opt-in
- Doc: Update README
- UI: Make disassambler a feature opt-in

Those are the 3 different perf with no feature, logger and disassembler
flamegraph

flamegraph1

flamegraph0


AlessioC31 (approved):
Nice, thanks

PR #217, opened 2023-12-30
Fix `LoadAddress`

AlessioC31:
In the `LoadAddress` Thumb instruction we were setting bit 0 to 0 while we should set bit 1 to zero, according to documentation.

I also did some small fixes around.


guerinoni (approved):
Amazing

PR #220, opened 2024-01-03
Fix blockmemory

AlessioC31:
* Thumb: Correctly handles base in register list for multiple load/store
     This commit fixes thumb tests 230,231,232.
* Thumb: Correctly emulates undefined behavior when register list is empty in block load/store
    This should never happen but test rom has a test for it.
    This commit fixes tests 227 and 229.


guerinoni (approved):
Nice found!

PR #222, opened 2024-02-16
Support BG2

AlessioC31:
* Moved registers and memory of LCD into appropriate structs
* Added basic support for BG2
* Fixed interrupt handler

Now if we run the Thumb tests we have:

image

🎉


guerinoni (approved):
This looks really good, it would be nice to add some reference in readme or in comment where those things are written but we can do this later aslo

PR #224, opened 2025-10-21
Fix `MultipleLoadStore` ASM

### LINKED ISSUE #190: Fix `MultipleLoadStore` ASM
Now we have `{R0, R1, }` in the register list (I mean at the end we would have an additional comma).

Also, to be consistent with the datasheet we should use `-` to separate registers.

_Originally posted by @AlessioC31 in https://github.com/RIP-Comm/clementine/pull/189#discussion_r1194734240_

raphaelts3:
This PR aims to make the improvements suggested on https://github.com/RIP-Comm/clementine/issues/190 by changing the register strings in the disassembler area from `{R0, R1, }` to `{R0-R1-R2...}`.

If approved, this PR closes https://github.com/RIP-Comm/clementine/issues/190.

### DIFF
--- a/emu/src/cpu/arm/instructions.rs
+++ b/emu/src/cpu/arm/instructions.rs
@@ -399,12 +399,13 @@ impl ArmModeInstruction {
                     Indexing::Post => "A",
                 };
 
-                let mut registers = String::new();
+                let mut regs = Vec::new();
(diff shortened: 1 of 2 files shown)


guerinoni (approved):
Thanks

